PR #510, opened 2026-03-23
fix: enforce Prettier formatting across all automated workflows

clarkemoyer:
## Summary

Three-layer defense against unformatted commits from AI agents and GitHub Actions:

- **Layer 1 — Patch known gap:** Add Node.js setup, `npm ci`, and `prettier --write` to `qualtrics-metrics-update.yml` before `peter-evans/create-pull-request` (mirrors the fix from #508 for `ga-report.yml`). Also updates automerge condition to trigger on both `created` and `updated`.

- **Layer 2 — Reusable composite action:** New `.github/actions/format-and-pr/action.yml` that bundles Node setup, npm ci, Prettier formatting, create-pull-request, and automerge into a single step. Future workflows can use this instead of implementing the pattern manually.

- **Layer 3 — Self-healing auto-fix:** New `prettier-autofix.yml` workflow that monitors CI failures on PR branches. When CI fails, it checks if Prettier is the cause and auto-pushes a formatting fix commit to the PR branch — so AI agents and automated tools that bypass pre-commit hooks get their code fixed automatically.

- **CLAUDE.md update:** Adds guidance for building Prettier-compliant workflows, documenting the composite action usage pattern and the safety net.

## Context

PR #435 (GA impact stats) failed CI because `generate-report.ts` wrote JSON without a trailing newline. We fixed that in #508 but identified `qualtrics-metrics-update.yml` as having the same gap. This PR systematically closes the problem repo-wide.

## Test plan

- [ ] CI passes (format check, lint, tests, build, E2E)
- [ ] Verify `qualtrics-metrics-update.yml` has Prettier step before commit
- [ ] Verify composite action YAML is valid
- [ ] Verify `prettier-autofix.yml` workflow syntax is valid (CodeQL will check)
- [ ] Trigger `Update Qualtrics Survey Metrics` workflow to confirm end-to-end

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -329,7 +329,7 @@ When you push to a PR, GitHub Actions runs:
 **Use the reusable composite action** for workflows that create data-update PRs:
 
 ```yaml
-# After downloading/generating data files:
+# Requires actions/checkout before this step:
 - uses: ./.github/actions/format-and-pr
   with:
     token: ${{ secrets.COPILOT_MCP_GITHUB_PERSONAL_ACCESS_TOKEN }}


clarkemoyer:
**Copilot Review Cycle (Round 1/7):** No fixes pushed within timeout. 3 comment(s) may need manual fixes.

PR #535, opened 2026-03-25
feat: DESTRUCTIVE — personalized rejection with Prolific categories and messages

clarkemoyer:
## Summary

> **DESTRUCTIVE OPERATION** — rejected participants will NOT be paid.

Reworks the rejection workflow from bulk-reject (no message) to **individual rejections with personalized messages matching the existing spreadsheet template**.

### What each participant sees

**Prolific categories checked:**
- "Failed attention check questions" (always)
- "Finished the study too quickly" (if Speed_Flag=1)
- "Other" (always, carries the detailed message)

**Message template:**
> Hi, thank you for participating in our Technology Adoption Barriers Survey. Unfortunately, your submission has been rejected because it was completed in X.X minutes (below our 5-minute minimum) and 3 of 3 embedded attention checks were answered incorrectly. These checks are designed to confirm that respondents are reading each question carefully. If you believe this is an error, please reply to this message with any questions.

### How it works

1. Exports fresh disposition data from Qualtrics
2. Filters for IRI_Fail_Count == 3
3. Looks up submission IDs from participant IDs via Prolific API
4. Rejects each submission individually with categories + personalized message
5. Logs every PID, message, and API response

### Safety guards (unchanged)
- Manual trigger only, dry-run default, REJECT confirmation string, dual validation, per-PID audit trail

## Test plan
- [x] 287 tests pass (6 new)
- [ ] Dry-run to preview all 30 rejection messages
- [ ] Verify messages match spreadsheet format

### DIFF
--- a/__tests__/lib/prolific-api.test.ts
+++ b/__tests__/lib/prolific-api.test.ts
@@ -12,6 +12,9 @@ import {
   exportSubmissionsCSV,
   bulkApproveSubmissions,
   bulkRejectSubmissions,
+  rejectSubmission,
+  getSubmissionIdsByParticipant,
+  REJECTION_CATEGORIES,
   ProlificApiErrorClass,
   type Study,
   type Submission,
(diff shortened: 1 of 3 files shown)


clarkemoyer:

1. **`scripts/reject-failed-iri-submissions.ts`**: The live rejection loop awaits each `rejectSubmission` without per-record error handling. If one rejection fails (network/API validation), the script will throw and stop, potentially leaving a partially rejected set with no structured summary of which PIDs succeeded vs failed. Consider wrapping each call in `try/catch`, recording successes/failures, and continuing (or at least emitting a final summary before rethrowing) so operators can safely reconcile the study state.

2. **`scripts/reject-failed-iri-submissions.ts`**: The failure Step Summary was simplified to “Check the job logs for error details.” Because this script performs destructive actions, the summary should explicitly warn that the run may have partially rejected some submissions before failing (depending on where the error occurred) and direct operators to the per-PID logs to reconcile state.
```suggestion
  appendGithubStepSummary(
    [
      '---',
      '## REJECTION FAILED',
      '',
      '> **WARNING:** This is a destructive script. Depending on where the error occurred, some submissions may have been rejected before the failure.',
      '> Please review the per-participant (PID) log entries in the job output to determine which submissions were actually rejected and reconcile state.',
      '',
      'See the job logs above for full error details.',
      '',
    ].join('\n')
  )
```

3. **`src/lib/prolific-api.ts`**: `getSubmissionIdsByParticipant` only inspects the first page returned by `listStudySubmissions`. Since `listStudySubmissions` returns a `PaginatedResponse` (with `meta.next`), this will miss submission IDs when a study has more submissions than the first page, causing false “not found” skips in the rejection script. Consider iterating pages until `meta.next` is null (or request a sufficiently large `page_size` if supported) before building the participant→submission map.

4. **`src/lib/prolific-api.ts`**: On success, `rejectSubmission` reads `response.text()` but only logs when the body is non-empty, and even then it logs only the status (not the response body). This makes per-submission audit logs inconsistent with `bulk-approve`/`bulk-reject` (which always log status and body/empty-body) and can result in no log line for a successful rejection with an empty response body. Consider always logging at least the status, and include the response body (or explicitly note empty body) for traceability.
```suggestion
    console.log(`  Rejected ${submissionId}: ${response.status} - ${responseBody}`)
  } else {
    console.log(`  Rejected ${submissionId}: ${response.status} (empty response body)`)
```

5. **`scripts/reject-failed-iri-submissions.ts`**: `DispositionRow` is imported as a type but never used. This will likely fail CI linting/TypeScript checks for unused imports. Remove the unused type import (or use it for `RejectionRecord` if that was the intent).
```suggestion
import { parseCsvLine } from '../src/lib/disposition'
```

Issue #544, opened 2026-03-25
Prolific API: Integrate AUTH CHECK (LLM + Bots) into disposition pipeline


clarkemoyer:
## Overview

Prolific provides two built-in authenticity checks per submission that we are NOT currently using:

| Check | Values | What it detects |
|---|---|---|
| **AUTH CHECK: LLM** | High / Mixed / Low | AI-generated responses |
| **AUTH CHECK: BOTS** | High / Mixed / Low | Automated/bot submissions |

These are independent from our Qualtrics-based checks (reCAPTCHA, IRI, speed, straightlining) and should be incorporated into the disposition triage.

## Current Gap

Our `Submission` interface in `prolific-api.ts` doesn't include these fields. The Prolific API likely returns them on the submission object but we're dropping them.

### 1. Update Prolific API types
Add auth check fields to the `Submission` interface:
```typescript
interface Submission {
  // ... existing fields
  auth_check_llm?: string     // "High" | "Mixed" | "Low"
  auth_check_bots?: string    // "High" | "Mixed" | "Low"
}
```

### 2. Fetch auth check data during triage
After running the Qualtrics-based triage, cross-reference with Prolific submission data to enrich the disposition CSV with:
- `Auth_LLM` (High/Mixed/Low)
- `Auth_Bots` (High/Mixed/Low)
- `Auth_Flag` (1 if either is not "High")

### 3. Update disposition waterfall
Add a new step for Prolific authenticity flags:
- **AUTH CHECK: Low on either** → `FLAG-AUTH-LOW` (manual review or auto-exclude)
- **AUTH CHECK: Mixed on either** → `FLAG-AUTH-MIXED` (manual review)

### 4. Waterfall placement
These should be checked EARLY in the waterfall (before IRI checks) since a bot/LLM response invalidates all other quality signals:

```
Step 0: INCOMPLETE
Step 0.5: FLAG-AUTH-LOW (Prolific LLM or Bots = Low) → AUTO-EXCLUDE?
Step 0.6: FLAG-AUTH-MIXED (Prolific LLM or Bots = Mixed) → manual review
Step 1: AUTO-EXCLUDE (IRI failures)
... rest of waterfall
```

### 5. Integration approach
The disposition pipeline currently only reads from Qualtrics. To add Prolific auth checks:

**Option A:** Add a step in the workflow that fetches submissions from Prolific, joins by PROLIFIC_PID, and enriches the disposition CSV before the triage step.

[...]

PR #549, opened 2026-03-25
feat: integrate Prolific AUTH CHECK (LLM + Bots) into disposition pipeline

clarkemoyer:
Adds Auth_LLM, Auth_Bots, Auth_Flag to DispositionRow. Early waterfall steps for FLAG-AUTH-FAIL and FLAG-AUTH-MIXED. Enrichment script for cross-referencing Prolific auth data. Implements #544.

### DIFF
--- a/__tests__/lib/prolific-api.test.ts
+++ b/__tests__/lib/prolific-api.test.ts
@@ -12,9 +12,17 @@ import {
   exportSubmissionsCSV,
   bulkApproveSubmissions,
   bulkRejectSubmissions,
+  rejectSubmission,
+  getSubmissionIdsByParticipant,
+  sendMessage,
+  listStudyMessages,
+  getMessageThread,
+  REJECTION_CATEGORIES,
   ProlificApiErrorClass,
   type Study,
   type Submission,
+  type Message,
+  type MessageThread,
   type PaginatedResponse,
 } from '../../src/lib/prolific-api'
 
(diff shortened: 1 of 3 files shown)


clarkemoyer:

1. **`src/lib/prolific-api.ts`**: This PR adds Prolific AUTH CHECK fields but leaves the exact API field names/value enums as a TODO. Given the feature depends on these names, consider making the client resilient by supporting the expected names from the issue (e.g., auth_check_llm/auth_check_bots) and/or adding a normalization layer (similar to normalizeStudy) so enrichment doesn’t silently produce empty auth values.

2. **`src/lib/disposition.ts`**: Auth_LLM/Auth_Bots/Auth_Flag are currently hard-coded to empty/0 for every row, so the new FLAG-AUTH-* waterfall steps can never trigger and Auth_Flag will always be 0. If the Qualtrics CSV is pre-enriched with these columns, triageCsv should read them from headers when present (and compute Auth_Flag accordingly) instead of overwriting them.
```suggestion
    // Auth check fields: prefer any pre-enriched CSV columns if present,
    // otherwise fall back to empty/0 for pipelines that enrich later.
    const authLlmIdx = headers.indexOf('Auth_LLM')
    const authBotsIdx = headers.indexOf('Auth_Bots')
    const authFlagIdx = headers.indexOf('Auth_Flag')

    const authLlm = authLlmIdx >= 0 ? (fields[authLlmIdx] ?? '').trim() : ''
    const authBots = authBotsIdx >= 0 ? (fields[authBotsIdx] ?? '').trim() : ''

    let authFlag: 0 | 1
    if (authFlagIdx >= 0) {
      const rawAuthFlag = (fields[authFlagIdx] ?? '').trim()
      if (rawAuthFlag === '') {
        // If the flag column exists but is blank, derive it from the auth fields.
        authFlag = authLlm || authBots ? 1 : 0
      } else {
        const normalized = rawAuthFlag.toLowerCase()
        authFlag =
          rawAuthFlag === '1' ||
          normalized === 'true' ||
          normalized === 'yes'
            ? 1
            : 0
      }
    } else {
      // No Auth_Flag column present; compute from any non-empty auth signals.
      authFlag = authLlm || authBots ? 1 : 0
    }
```

3. **`scripts/fetch-prolific-auth-checks.ts`**: When OUTPUT_PATH is not set, this script writes CSV to stdout but also prints progress logs via `console.log`, which will corrupt the CSV stream (e.g., when piping/redirecting). Route non-CSV logs to stderr (console.error) or suppress logs when outputting to stdout.

PR #580, opened 2026-03-27
feat: add SPEED_IRI message type for borderline auto-exclude participants

clarkemoyer:
Adds AUTO-EXCLUDE:SPEED_IRI as a new messaging option. These are participants who completed under 5 min with 1 IRI failure — borderline cases that deserve inquiry before rejection. Message asks for background, survey thoughts, and speed explanation.

### DIFF
--- a/.github/workflows/check-participant.yml
+++ b/.github/workflows/check-participant.yml
@@ -11,6 +11,7 @@ on:
           - single
           - all-replies
           - send-thank-you
+          - unreject
         default: 'single'
       pid:
         description: 'Prolific Participant ID (required for single mode)'
@@ -36,6 +37,15 @@ jobs:
           node-version: '20'
           cache: 'npm'
       - run: npm ci
+
+      - name: Unreject submissions
+        if: inputs.mode == 'unreject'
+        env:
+          PROLIFIC_API_TOKEN: ${{ secrets.TABS_PROLIFIC_TOKEN }}
+          STUDY_ID: ${{ inputs.study_id || vars.PROLIFIC_STUDY_ID }}
+          PID_LIST: ${{ inputs.pid }}
+          DRY_RUN: 'false'
+        run: npx --no-install tsx scripts/unreject-submissions.ts
 
       - name: Send thank-you messages
         if: inputs.mode == 'send-thank-you'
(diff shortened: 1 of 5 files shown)


clarkemoyer:

1. **`scripts/message-flagged-submissions.ts`**: `getMessageSignature()` is used with `listUserMessages()`, which returns messages sent **to or from** the participant. Using a question prompt as the signature ("What is your professional background and role") increases the chance a participant reply contains the same text (e.g., copying the numbered questions), causing a false "already sent" match and skipping messaging. Use a declarative phrase from the template that participants are unlikely to quote, and/or restrict the signature check to researcher-sent messages (e.g., filter by `sender_id`).
```suggestion
      return 'automatic exclusion based on your survey speed and attention check responses'
```

2. **`scripts/message-flagged-submissions.ts`**: The new message template reads as a run-on sentence ("...completed in X minutes and one of three...") because there’s no comma after "minutes". Add the comma (or adjust punctuation) to keep the message grammatically consistent with the other templates.
```suggestion
        `We are reviewing your submission because it was completed in ${minutes} minutes,`,
```

3. **`scripts/message-flagged-submissions.ts`**: `IRI_Fail_Count` and `Speed_Flag` are now required columns for *all* runs because `col()` throws when headers are missing. These fields are only needed for the `AUTO-EXCLUDE:SPEED_IRI` compound filter; requiring them unconditionally breaks messaging for other dispositions when using older/alternate CSV exports. Consider looking up these column indexes only when `DISPOSITION_FILTER === 'AUTO-EXCLUDE:SPEED_IRI'` (or make them optional with a clear error only when that filter is selected).
```suggestion

  let iriFailIdx = -1
  let speedIdx = -1

  if (dispositionFilter === 'AUTO-EXCLUDE:SPEED_IRI') {
    iriFailIdx = headers.indexOf('IRI_Fail_Count')
    speedIdx = headers.indexOf('Speed_Flag')

    if (iriFailIdx === -1 || speedIdx === -1) {
      console.error(
        'Error: CSV must include "IRI_Fail_Count" and "Speed_Flag" columns when DISPOSITION_FILTER is "AUTO-EXCLUDE:SPEED_IRI"',
      )
      process.exit(1)
    }
  }
```

PR #600, opened 2026-03-29
feat: add V2 psychometric analysis & descriptive stats scripts

clarkemoyer:
## Summary

- Add reproducible analysis pipeline for TABS V2 data (N=277 total, N=102 clean)
- `tabs_v2_analysis.py`: descriptive stats, cross-tabs, effect sizes, correlations, Cronbach's α, and sensitivity analysis across 3 sample cuts (Clean/Relaxed/All V2)
- `tabs_v2_psychometrics.py`: convergent/discriminant validity (MTMM-style), IRI effectiveness validation, response bias detection, order/fatigue effects, response entropy
- No raw data embedded — scripts accept a Qualtrics CSV export as their only argument

## Key Findings from Analysis

- B–R correlation: r = −0.50 (clean) → −0.15 (unfiltered), validating IRI quality filtering
- 95% convergent validity (41/43 items load on own scale)
- Cronbach's α: .864–.914 across all scales
- Largest effect: large-org barrier premium d = 0.95
- Zero straightlining, zero acquiescence bias in clean sample
- Item rankings stable across all 3 sample cuts

## Test plan

- [ ] Run `python scripts/analysis/tabs_v2_analysis.py <csv_path>` and verify output matches report
- [ ] Run `python scripts/analysis/tabs_v2_psychometrics.py <csv_path>` and verify output matches report
- [ ] Confirm no raw data is embedded in any committed file

### DIFF
--- a/scripts/analysis/tabs_v2_analysis.py
+++ b/scripts/analysis/tabs_v2_analysis.py
@@ -19,7 +19,7 @@ Author: Clarke Moyer, Penn State Smeal DBA
 import csv
 import math
 import sys
-from collections import Counter, defaultdict
+from collections import Counter
 
 # ─────────────────────────────────────────────────────────────
 # Configuration
@@ -106,6 +106,7 @@ NONTECH_TITLES = {'CEO', 'CFO', 'COO', 'CHRO', 'CMO', 'CSO', 'CRO'}
 
 def mean_sd(values):
     """Compute mean and sample standard deviation."""
+    values = [v for v in values if v is not None]
     n = len(values)
     if n == 0:
         return (None, None)
@@ -239,6 +240,8 @@ def person_means(rows, cols, scale, idx):
         vals = [v for v in vals if v is not None]
         if vals:
             out.append(sum(vals) / len(vals))
+        else:
+            out.append(None)
     return out
 
 
@@ -323,6 +326,9 @@ def print_disposition(v2, clean, relaxed, all_v2, idx):
 
 def print_demographics(rows, idx, label="Clean"):
     """Print demographic breakdown."""
+    if not rows:
+        print("  No data")
+        return
     print(f"\n{'=' * 78}")
     print(f"  DEMOGRAPHICS ({label} N={len(rows)})")
     print("=" * 78)
@@ -376,11 +382,15 @@ def print_item_rankings(rows, idx):
         print(f"\n  {'Rank':>4} {'Item':<38} {'M':>6} {'SD':>6} {'n':>4}")
         print(f"  {'─' * 4} {'─' * 38} {'─' * 6} {'─' * 6} {'─' * 4}")
         for rank, (name, m, sd, n) in enumerate(stats, 1):
-            print(f"  {rank:>4} {name:<38} {m:.2f}  {sd:.2f}  {n:>3}")
+            m_str = f"{m:.2f}" if m is not None else "NA"
+            sd_str = f"{sd:.2f}" if sd is not None else "NA"
+            print(f"  {rank:>4} {name:<38} {m_str:>6}  {sd_str:>6}  {n:>3}")
 
         pm = person_means(rows, cols, sc, idx)
         gm, gsd = mean_sd(pm)
-        print(f"\n  Grand Mean: {gm:.2f} (SD={gsd:.2f})")
+        gm_str = f"{gm:.2f}" if gm is not None else "NA"
+        gsd_str = f"{gsd:.2f}" if gsd is not None else "NA"
+        print(f"\n  Grand Mean: {gm_str} (SD={gsd_str})")
 
 
 def print_correlations_and_reliability(rows, idx):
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. **`scripts/analysis/tabs_v2_psychometrics.py`**: `defaultdict` is imported but not used, which will trigger linting warnings in many environments. Please remove unused imports (or use them if intended).
```suggestion
from collections import Counter
```

2. **`scripts/analysis/tabs_v2_analysis.py`**: In `print_effect_sizes`, `cohens_d()` can return `None` and `mean_sd()` can return `(None, None)` when groups are too small; formatting `d={d:+.2f}` / `Tech={tm:.2f}` will then crash. Also, the `size = ...` expression treats `d=0.0` as falsy and will label it "large". Use explicit `is not None` checks and print "N/A" when group sizes are insufficient.

3. **`scripts/analysis/tabs_v2_analysis.py`**: `print_item_rankings` formats item means/SDs and the grand mean with `:.2f`, but `mean_sd([])` returns `(None, None)` when an item has no scorable responses (e.g., missing data or changed Qualtrics labels). Formatting `None` will raise `TypeError`. Consider skipping items with no data or printing "N/A" for those rows.

4. **`scripts/analysis/tabs_v2_analysis.py`**: `defaultdict` is imported but not used. Please remove unused imports to keep the scripts clean and avoid lint failures in environments that enforce them.
```suggestion
from collections import Counter
```

5. **`scripts/analysis/tabs_v2_psychometrics.py`**: `section_scale_discrimination` assumes the within/between average correlations are always present and non-zero (`{bw:.3f}`, `bw / max(...)`). If any average is `None` (no valid item pairs) or the max-between is 0, this will raise `TypeError`/`ZeroDivisionError`. Add guards (e.g., print "N/A" and skip ratios when inputs are missing/zero).

6. **`scripts/analysis/tabs_v2_psychometrics.py`**: `section_iri_effectiveness` divides by `len(speed_ok)` in multiple places. If no rows meet the duration threshold, this will throw `ZeroDivisionError`. Add an early return/branch when `speed_ok` is empty so the script can still complete with a clear message.

7. **`scripts/analysis/tabs_v2_psychometrics.py`**: In `section_order_fatigue`, `q_size = len(durations) // 4` can be 0 when `len(durations) < 4`, which makes `durations[q_size]` raise `IndexError` and also makes `slow = durations[-q_size:]` unexpectedly include all rows (because `-0` is `0`). Guard for small samples (e.g., require at least 4 durations, or choose a different split).

8. **`scripts/analysis/tabs_v2_psychometrics.py`**: `get_duration` uses a bare `except:` which will swallow unexpected errors (including programming bugs) and can hide malformed headers/rows. Please catch specific exceptions (e.g., `KeyError`/`ValueError`/`IndexError`) and consider failing fast with a helpful message if the duration column is missing.
```suggestion
    duration_key = 'Duration (in seconds)'
    if duration_key not in idx:
        raise KeyError(
            "Required duration column 'Duration (in seconds)' is missing from header index."
        )
    col_index = idx[duration_key]
    try:
        return int(row[col_index])
    except (ValueError, IndexError):
        # Non-numeric value or missing column value for this row; treat as no duration
        return None
```

9. **`scripts/analysis/tabs_v2_psychometrics.py`**: The comment says "Import shared configuration and helpers from analysis module", but this file actually duplicates the constants and helpers instead of importing them. Either import from a shared module (to prevent drift between scripts) or update the comment to reflect the current approach.

10. **`scripts/analysis/tabs_v2_analysis.py`**: `print_demographics` divides by `len(rows)` in several percentage calculations. If the Clean sample is empty (e.g., due to stricter filters or a wrong V2_START), this will raise `ZeroDivisionError`. Add an early check for `if not rows:` to print a message and return.

11. **`scripts/analysis/tabs_v2_analysis.py`**: Several report sections assume computed stats are always numeric and format them with `:.2f`/`:.3f`/`:+.2f`, but helper functions can return `None` (e.g., `pearson_r` when n<3, `cronbach_alpha` when n<3 complete cases, `skewness`/`kurtosis_excess` for small samples). As written, lines like `r = {br:.3f}` and `skew={sk:+.2f}` will raise `TypeError`. Please format `None` as "N/A" and/or guard these prints.

12. **`scripts/analysis/tabs_v2_psychometrics.py`**: `section_convergent_discriminant` prints correlations using float format specifiers, but `pearson_r()` can return `None` (e.g., n<3 or zero variance). Formatting `None` with `+7.3f` will raise a `TypeError` and stop the script. Consider formatting `None` as "N/A" (and updating the CONV/CROSS-LOAD logic to explicitly check `is not None`).

clarkemoyer:

1. **`scripts/analysis/README.md`**: The Markdown table under “## Scripts” uses double leading pipes (`|| ...`), which renders as an extra empty column in most Markdown renderers. Use a standard table row format with single leading/trailing pipes so the table displays correctly on GitHub.

2. **`scripts/analysis/tabs_v2_psychometrics.py`**: `section_scale_discrimination` formats `bw/rw/mw/br_b/...` with `:.3f` even when `avg_within/avg_between` return `None` (e.g., empty/too-small clean sample or zero-variance items). This will raise a `TypeError` and stop the script. Consider guarding for `None` before formatting (print `N/A`) or early-return when correlations can’t be computed.

3. **`scripts/analysis/tabs_v2_psychometrics.py`**: In the acquiescence section, `person_grand` can be empty even when `clean` is non-empty (e.g., if all scale items are missing/unscorable). In that case `mean_sd(person_grand)` returns `(None, None)` and the `:.2f` formatting will crash; additionally, the percentages are divided by `len(clean)` rather than `len(person_grand)`, which can misstate rates. Guard for empty `person_grand` and use the correct denominator.
```suggestion
            if v:
                all_vals.append(v)
        for c in READINESS_COLS:
            v = score(r, c, READINESS_SCALE, idx)
            if v:
                all_vals.append(v)
        for c in MATURITY_COLS:
            v = score(r, c, MATURITY_SCALE, idx)
            if v:
                all_vals.append(v)
        if all_vals:
            person_grand.append(sum(all_vals) / len(all_vals))

    if person_grand:
        pm_m, pm_s = mean_sd(person_grand)
        print(f"    Grand mean of all-item person means: {pm_m:.2f} (SD={pm_s:.2f})")
        print(f"    Theoretical midpoint: 3.00")
        high = sum(1 for v in person_grand if v >= 4.0)
        low = sum(1 for v in person_grand if v <= 2.0)
        denom = len(person_grand)
        print(f"    Mean >= 4.0 (acquiescence): {high} ({high / denom * 100:.1f}%)")
        print(f"    Mean <= 2.0 (naysaying):    {low} ({low / denom * 100:.1f}%)")
    else:
        print("    No scorable responses for acquiescence bias (person-level means are empty).")
```

4. **`scripts/analysis/tabs_v2_psychometrics.py`**: In the extreme response style section, `ers_m/ers_s` can be `(None, None)` if `ers` is empty, and the `:.2f` formatting will crash. Also, the high-ERS percentage is computed over `len(clean)` rather than `len(ers)`, which can be inaccurate if some respondents have no scorable items. Add an empty-check for `ers` and use the appropriate denominator.
```suggestion
    if not ers:
        print("    No scorable responses for ERS (extreme response style).")
    else:
        ers_m, ers_s = mean_sd(ers)
        high_ers = sum(1 for v in ers if v > 0.60)
        print(f"    Mean extreme proportion: {ers_m:.2f} (SD={ers_s:.2f})")
        print(f"    Expected under uniform: 0.40")
        print(f"    High ERS (>60%): {high_ers} ({high_ers / len(ers) * 100:.1f}%)")
```

5. **`scripts/analysis/tabs_v2_psychometrics.py`**: `section_order_fatigue` appends `sd` values that may be `None` (when an item has no scorable responses), then calls `sum(f_sds)` / `sum(s_sds)`, which will raise `TypeError`. Filter out `None` SDs (and handle the case where all SDs are missing) before averaging/printing.
```suggestion
            if sd is not None:
                f_sds.append(sd)
        for c in second:
            vals = [score(r, c, sc, idx) for r in clean]
            vals = [v for v in vals if v is not None]
            _, sd = mean_sd(vals)
            if sd is not None:
                s_sds.append(sd)

        if f_sds and s_sds:
            avg_f = sum(f_sds) / len(f_sds)
            avg_s = sum(s_sds) / len(s_sds)
            delta = avg_s - avg_f
            interp = (
                "stable"
                if abs(delta) < 0.05
                else "possible fatigue"
                if delta > 0
                else "possible engagement increase"
            )
            print(f"    {label:12s}: first={avg_f:.3f}, second={avg_s:.3f}, delta={delta:+.3f} ({interp})")
        else:
            print(f"    {label:12s}: first=N/A, second=N/A, delta=N/A (insufficient data)")
```

6. **`scripts/analysis/tabs_v2_analysis.py`**: `ROLE_MAP` omits the CISO option that’s listed in the survey metadata (e.g., concept-mapping-simple.json describes a CISO response option). CISO respondents will be labeled `Unknown`, which will skew demographic breakdowns and any tech/non-tech comparisons. Add the missing mapping (and decide whether CISO belongs in `TECH_TITLES`/`NONTECH_TITLES`).
```suggestion
    'CISO (e.g., Chief Information Security Officer)': 'CISO',
    'Other (please specify)': 'Other'
}
TECH_TITLES = {'CIO', 'CTO', 'CISO'}
```

7. **`scripts/analysis/tabs_v2_analysis.py`**: `cohens_d` uses `n1 = len(g1)` / `n2 = len(g2)` even though `mean_sd()` drops `None` values. If `g1/g2` contain `None` entries, the pooled SD (and d) will be computed with sample sizes that don’t match the means/SDs, producing incorrect effect sizes. Filter out `None` values first and use the filtered lengths consistently.

8. **`scripts/analysis/tabs_v2_analysis.py`**: `get_duration` doesn’t catch `IndexError`, so a malformed/short CSV row could crash the script when indexing `row[idx['Duration (in seconds)']]`. The psychometrics script already catches `IndexError`; consider matching that here for robustness.
```suggestion
    except (ValueError, KeyError, IndexError):
```

clarkemoyer:
**Copilot Review Cycle (Round 4/7):** No fixes pushed within timeout. 8 comment(s) may need manual fixes.

PR #602, opened 2026-03-29
fix: revert broken per-PID status check to batch fetch

clarkemoyer:
getSubmission(studyId, subId) returns 404 for all PIDs. Reverted to batch listStudySubmissions fetched right before the send loop. Status guard still works.


clarkemoyer:
**Copilot Review Cycle (Round 1/7):** Timed out waiting for Copilot review. You can re-trigger manually.

PR #614, opened 2026-03-30
fix: address Copilot review comments on daily automation workflow

clarkemoyer:
## Summary

Fixes Copilot review comments from PR #612 and hardens the daily disposition processing workflow:

- Add export column verification step (`check-export-headers.ts` now exits non-zero on missing columns)
- Split dashboard generation (prolific-prod) from PR creation (copilot env) via artifact handoff
- Add `set -euo pipefail` to all shell steps so `| tee` doesn't hide script failures
- Move daily report generation to external `scripts/build-daily-report.sh` (avoids YAML/markdown conflicts)
- Create a new GitHub issue per daily run with full triage, approval, and messaging details + deltas
- Add `issues: write` permission for issue creation
- Fix delta computation to use consistent `totalResponses` field

## Test plan

- [ ] Trigger workflow manually via `workflow_dispatch`
- [ ] Verify export column check catches missing headers
- [ ] Verify daily report issue is created with correct markdown rendering
- [ ] Verify approve/message/dashboard steps run successfully

### DIFF
--- a/.github/workflows/daily-disposition-processing.yml
+++ b/.github/workflows/daily-disposition-processing.yml
@@ -155,6 +155,8 @@ jobs:
           DISPOSITION_FILTER: ${{ matrix.disposition }}
           DRY_RUN: 'false'
 
+      # Note: message-output.txt contains PIDs for operational debugging.
+      # Artifact has 1-day retention; same data appears in workflow logs.
       - uses: actions/upload-artifact@v7
         if: always()
         with:
--- a/scripts/build-daily-report.sh
+++ b/scripts/build-daily-report.sh
@@ -7,6 +7,14 @@ set -euo pipefail
 # Required env vars: GH_TOKEN, RUN_URL, ARTIFACTS_DIR, PREV_FILE
 # Optional: TRIAGE_RESULT, APPROVE_RESULT_STATUS, MESSAGE_RESULT, DASHBOARD_RESULT
 
+# Validate required env vars
+for var in GH_TOKEN RUN_URL; do
+  if [ -z "${!var:-}" ]; then
+    echo "Error: required env var $var is not set" >&2
+    exit 1
+  fi
+done
+
 DATE=$(date -u +"%Y-%m-%d")
 ARTIFACTS_DIR="${ARTIFACTS_DIR:-.}"
 PREV_FILE="${PREV_FILE:-src/data/disposition-summary.json}"


clarkemoyer:

1. **`.github/workflows/daily-disposition-processing.yml`**: The `triage-total.txt` extraction will often produce an empty file when the "Total responses" line is missing because the `grep | sed` pipeline’s exit status is from `sed` (0 even when `grep` finds nothing). That means the `|| echo "0"` fallback won’t run, and the daily report may show a blank triage total. Fix by enabling `set -euo pipefail` in this step and restructuring the extraction to explicitly handle the no-match case (e.g., `grep -m1 ... || echo '**Total responses:** 0'` before `sed`, or avoid a pipeline).
```suggestion
          set -euo pipefail
          # Extract total (handle case where no responses exist)
          total_line="$(grep -m1 '\*\*Total responses:\*\*' "${{ runner.temp }}/triage-output.txt" || echo '**Total responses:** 0')"
          printf '%s\n' "$total_line" | sed 's/.*\*\*Total responses:\*\* //' > "${{ runner.temp }}/triage-total.txt"
```

2. **`scripts/build-daily-report.sh`**: The header comment lists optional env vars as `TRIAGE_RESULT, APPROVE_RESULT, MESSAGE_RESULT, DASHBOARD_RESULT`, but the script actually consumes `APPROVE_RESULT_STATUS` (and the workflow sets `APPROVE_RESULT_STATUS`). Update the comment so the documented env vars match what the script reads.
```suggestion
# Optional: TRIAGE_RESULT, APPROVE_RESULT_STATUS, MESSAGE_RESULT, DASHBOARD_RESULT
```

clarkemoyer:

1. **`.github/workflows/daily-disposition-processing.yml`**: The approve-metrics artifact upload won’t run if the previous “Approve CLEAN submissions” step fails (default behavior), which makes debugging harder and can leave the daily report without approve output. Consider adding `if: always()` (or at least `if: success() || failure()`) to ensure `approve-output.txt` is uploaded even on failures.

2. **`.github/workflows/daily-disposition-processing.yml`**: The message-metrics artifact upload step will be skipped if the message-sending step exits non-zero, which means failed runs won’t retain `message-output.txt` for that disposition and the daily report may miss the per-disposition row. Add `if: always()` to the upload step (and optionally `continue-on-error: true` for the upload) so logs/metrics are still captured on failures.

clarkemoyer:

1. **`.github/workflows/daily-disposition-processing.yml`**: The workflow now captures `scripts/message-flagged-submissions.ts` output via `tee` and uploads it as an artifact. That script logs participant PIDs (and the full message text) for every matched participant, so this change persists participant identifiers in both Actions logs and downloadable workflow artifacts. Consider producing/uploading a metrics-only file (counts) or redacting PIDs/message bodies before `tee`/artifact upload to avoid storing PII in CI artifacts.

2. **`scripts/build-daily-report.sh`**: With `set -euo pipefail`, missing required env vars (e.g., RUN_URL) will currently fail as an unhelpful “unbound variable” error at heredoc expansion time. Adding explicit checks with a clear error message for the documented required variables (RUN_URL/GH_TOKEN, etc.) would make failures easier to diagnose when this script is run outside the workflow or if env wiring regresses.

clarkemoyer:

1. **`scripts/build-daily-report.sh`**: The header comment says `ARTIFACTS_DIR` and `PREV_FILE` are required, but the script provides defaults for both and only validates `GH_TOKEN` and `RUN_URL`. Please update the comment to match the actual requirements (or validate those vars if they truly must be set) to avoid confusion for future workflow edits.

PR #616, opened 2026-03-30
feat: fix KMO/HTMT computations, add data audit script

clarkemoyer:
## Summary
- **Fixed KMO computation** in `tabs_v2_psychometrics.py` — off-diagonal sum of squares was always 0 for symmetric matrices (now produces KMO = 0.927 "Marvelous")
- **Fixed HTMT computation** — `corrwith()` matched columns by name across constructs with different column names, yielding NaN (now produces valid ratios: B-R 0.380, B-M 0.377, R-M 0.800, all below 0.85 threshold)
- **Added `tabs_v2_data_audit.py`** — comprehensive data quality audit covering disposition funnel, demographic imbalances, missing data patterns, response quality indicators, and temporal analysis
- **Updated README** with new script in the table and usage section

## Test plan
- [ ] Run `python scripts/analysis/tabs_v2_psychometrics.py <csv>` and verify KMO ≈ 0.927 and HTMT ratios are numeric (not NaN)
- [ ] Run `python scripts/analysis/tabs_v2_data_audit.py <csv>` and verify it produces the full 10-section audit output
- [ ] Verify all three scripts accept CSV path as `sys.argv[1]`


clarkemoyer:

1. **`scripts/analysis/tabs_v2_psychometrics.py`**: `load_data()` doesn’t appear to handle the standard Qualtrics 3-row header (column names, sub-labels, import IDs). `pd.read_csv()` will treat the 2nd/3rd header rows as data rows, but the code only drops one row (`df.iloc[1:]`), which will shift/contaminate all subsequent numeric encodings and validity statistics. Consider reading with `encoding='utf-8-sig'` and skipping the two non-data rows (e.g., `skiprows=[1,2]` while keeping the first row as headers) so parsing matches `scripts/analysis/README.md` and `tabs_v2_analysis.py`.
```suggestion
    # Read CSV with Qualtrics 3-row header:
    # row 0 = field names (used as headers), rows 1–2 = metadata (skipped)
    df = pd.read_csv(CSV_PATH, encoding='utf-8-sig', skiprows=[1, 2])
```

2. **`scripts/analysis/tabs_v2_psychometrics.py`**: The “Alpha if Item Deleted” improvement calculation is reversed: `improvement = alpha - alpha_new` makes the value positive when deleting the item *reduces* alpha, but the output/flag says it “IMPROVES”. This will mislabel items for deletion. Compute the delta as `alpha_new - alpha` (or invert the flag condition) so the “improves by ≥ .01” marker is only applied when alpha increases after deletion.

3. **`scripts/analysis/tabs_v2_psychometrics.py`**: `check_iri()` is defined but never used (no call sites in this script). Consider removing it to keep the analysis script smaller, or using it where you currently duplicate IRI-pass logic (e.g., when building `IRI_*_Pass` columns).
```suggestion

```

4. **`scripts/analysis/tabs_v2_psychometrics.py`**: `f_oneway` is imported but not used anywhere in this script. Removing unused imports helps keep dependencies clear and avoids confusion about intended analyses.
```suggestion
from scipy.stats import shapiro, ttest_ind, chi2_contingency
```

5. **`scripts/analysis/tabs_v2_psychometrics.py`**: In the “Readiness Sub-Construct Coverage” output, items are labeled with a `C` prefix (`(C{items})`) while the other constructs use `B`/`M`. If this is intended to be “Readiness” item numbering, consider using an `R` prefix (or another explicit label) to avoid confusing readers of the report.
```suggestion
        print(f"  {name}: {len(items)} items (R{items})")
```

6. **`scripts/analysis/tabs_v2_data_audit.py`**: On duration parse failures the script appends `0` seconds, which will skew min/percentiles and speed-tier counts (and may incorrectly increase “Very fast” responses). Instead, treat unparseable durations as missing (e.g., `np.nan`) and compute percentiles/tier counts on valid values while also reporting how many durations were missing/unparseable.
```suggestion
missing_durations = 0
for r in all_rows:
    try:
        durations.append(float(r[idx['Duration (in seconds)']]))
    except Exception:
        durations.append(np.nan)
        missing_durations += 1

dur_arr = np.array(durations, dtype=float)
valid_dur_arr = dur_arr[~np.isnan(dur_arr)]
n_valid = valid_dur_arr.size

print(f"\n  Duration distribution (all V2, N={N_all}, valid durations={n_valid}):")
print(f"    Missing/unparseable durations: {missing_durations}")
if n_valid > 0:
    print(f"    Min:    {valid_dur_arr.min():.0f}s ({valid_dur_arr.min()/60:.1f}m)")
    print(f"    P5:     {np.percentile(valid_dur_arr, 5):.0f}s")
    print(f"    P25:    {np.percentile(valid_dur_arr, 25):.0f}s ({np.percentile(valid_dur_arr, 25)/60:.1f}m)")
    print(f"    Median: {np.percentile(valid_dur_arr, 50):.0f}s ({np.percentile(valid_dur_arr, 50)/60:.1f}m)")
    print(f"    P75:    {np.percentile(valid_dur_arr, 75):.0f}s ({np.percentile(valid_dur_arr, 75)/60:.1f}m)")
    print(f"    P95:    {np.percentile(valid_dur_arr, 95):.0f}s ({np.percentile(valid_dur_arr, 95)/60:.1f}m)")
    print(f"    Max:    {valid_dur_arr.max():.0f}s ({valid_dur_arr.max()/60:.1f}m)")
else:
    print("    No valid duration values available to compute statistics.")
```

7. **`scripts/analysis/tabs_v2_data_audit.py`**: `clean` can be 0 for some datasets (e.g., small exports or stricter-than-expected filters). Many later sections divide by `clean`/`total_cells`, which will raise `ZeroDivisionError` and abort the audit. Consider adding an early guard after computing `clean` (and/or using safe percentage helpers) to print a clear message and exit or skip clean-sample sections when `clean == 0`.

8. **`scripts/analysis/tabs_v2_data_audit.py`**: The report text includes a placeholder “$X compensation” rather than the actual incentive amount. If the amount is known, consider filling it in; if it’s intentionally unknown, rephrase to avoid implying a missing value (e.g., “compensation amount”).
```suggestion
  c) Payment effect: monetary compensation may attract satisficers who rush
```

clarkemoyer:
**Copilot Review Cycle (Round 1/7):** No fixes pushed within timeout. 4 comment(s) may need manual fixes.

PR #621, opened 2026-03-31
chore: Verify Barriers and Align Visual Text (Closes #620)

### LINKED ISSUE #620: Verify Site Barriers Match Final Survey Verbatim
We need to audit the 24 barriers defined in \src/data/barriers.ts\ and confirm they match word-for-word against the text listed in the final Qualtrics survey (SV_bkMopd73A8fzfwO). The live site must perfectly reflect the questions asked to the participants to ensure alignment between public documentation and the collected data.

clarkemoyer:
This PR fixes the fetching script to correctly parse Qualtrics v3 API data, and adjusts the Barriers page UI to render the verbatim Qualtrics survey text in the header to ensure perfect stylistic alignment with the scientific instrument. Closes #620

### DIFF
--- a/__tests__/components/tabs/Barriers.test.tsx
+++ b/__tests__/components/tabs/Barriers.test.tsx
@@ -40,6 +40,7 @@ describe('Barriers component', () => {
 
       // The barrier with matching description should be visible
       expect(screen.getByText(firstBarrier.description)).toBeInTheDocument()
+      expect(screen.getByText(firstBarrier.name)).toBeInTheDocument()
     })
 
     it('should filter barriers by description when searching', () => {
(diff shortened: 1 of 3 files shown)


clarkemoyer:

1. **`scripts/fetch-qualtrics-questions.ts`**: `getSurveyQuestions()` returns the parsed `result` object from `makeApiRequest`, but this script still assumes multiple possible shapes without validating them. As written, `data.elements` is accessed without guarding against `data` being null/undefined, and the later loop still assumes legacy `QuestionID/QuestionText/Choices` field names (which will be undefined if the v3 `elements[]` objects are camelCased), causing runtime failures (the added workflow log shows the script still crashing). Add explicit shape detection (e.g., `Array.isArray(data?.elements)`), normalize v3 question objects into the expected `QualtricsQuestion` shape (or update `QualtricsSurveyDefinition`/`getSurveyQuestions` typings to match v3), and throw a clear error when the response schema is unexpected.

2. **`src/components/tabs/barriers/index.tsx`**: This change removes `barrier.name` from the rendered card entirely and uses `barrier.description` as the `<h3>` text. However, the search/filtering logic earlier in this component still matches on `barrier.name`, so users can get search hits for text that is never visible in the UI. Either render the name somewhere (even as secondary text) or update the search matching to align with what’s displayed.

clarkemoyer:
Fixed in latest commit. Re-requesting review.

PR #647, opened 2026-04-01
feat: improve concept mapping table readability + Excel download

clarkemoyer:
## Summary

Improves readability of the 57-row × 13-column concept mapping table and adds Excel download.

### Changes
- **Column widths**: Shrink sparse columns (RIS Citation, Variable Type, etc.) via `maxWidth` so content-heavy columns get more space
- **Min width**: Reduced from 120px to 100px; columns with explicit maxWidth override min-width
- **Excel download**: Primary download button generates proper `.xlsx` via `exceljs` (dynamically imported on click to avoid bundle bloat)
- **CSV**: Kept as secondary option
- **Filter-aware exports**: Both CSV and Excel respect active search/section filters
- **Security**: Replaced `xlsx@0.18.5` (CVE-2023-30533, CVE-2024-22363) with `exceljs` (no known CVEs)
- **Test**: Added test for Download Excel button

Closes #575

## Test plan

- [ ] View `/concept-mapping/simple` — verify table fits more columns without horizontal scroll
- [ ] Click "Download Excel" — verify `.xlsx` opens correctly in Excel/Sheets
- [ ] Filter to Section A, then download — verify only 9 rows in the file
- [ ] Click "CSV" — verify existing CSV export still works
- [ ] Test on mobile — verify horizontal scroll still works

### DIFF
--- a/src/components/tabs/concept-mapping/simple.tsx
+++ b/src/components/tabs/concept-mapping/simple.tsx
@@ -42,7 +42,7 @@ const EXPANDABLE_COLUMNS = new Set([
  * Column max-widths so sparse columns (e.g. RIS Citation, mostly N/A) shrink
  * and content-heavy columns get more space. Applied via inline style.
  */
-const COLUMN_MAX_WIDTHS: Record<string, string> = {
+const COLUMN_MAX_WIDTHS: Partial<Record<(typeof conceptMappingData.headers)[number], string>> = {
   'RIS Citation': '70px',
   'Source Link (URL/DOI)': '90px',
   'Variable Type': '100px',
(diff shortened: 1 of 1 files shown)


clarkemoyer:

1. **`src/components/tabs/concept-mapping/simple.tsx`**: The table still uses the Tailwind class `w-max` (`width: max-content`), which tends to force columns to expand to their content and can undermine the goal of shrinking sparse columns via `maxWidth` (it often results in more horizontal scrolling than necessary). Consider removing `w-max` and relying on `min-w-full` + the `overflow-x-auto` wrapper so the browser can actually shrink columns within the viewport.

clarkemoyer:
**Copilot Review Cycle (Round 5/7):** No fixes pushed within timeout. 1 comment(s) may need manual fixes.

Issue #574, opened 2026-03-26
Fix column with on all these pages to bet fit as many columns as possible.


clarkemoyer:
<img width="3436" height="1618" alt="Image" src="https://github.com/user-attachments/assets/054102de-1d71-42cb-8c8a-8cdedb019cee" />

PR #659, opened 2026-04-02
feat: add Excel/CSV downloads to all concept mapping views

clarkemoyer:
## Summary

Completes issue #574 scope and fixes the last known limitation:

### Downloads for all concept mapping views
- **New shared component**: `download-buttons.tsx` — reusable Excel/CSV export with dynamic exceljs import
- **Summary view**: Download buttons for the survey structure table
- **Complex view**: Bulk download buttons for all 57 items (respects search filter)

### Pagination fix for Prolific API
- `listStudySubmissions` now follows `meta.next` links to fetch all pages
- Adds generic `fetchAllPages` helper to `prolific-api.ts`
- Removes the TODO comment that flagged this limitation
- Prevents silent data loss if study grows beyond API page size

## Test plan

- [ ] Visit `/concept-mapping/summary` — verify Download Excel and CSV buttons
- [ ] Visit `/concept-mapping/complex` — verify bulk download buttons
- [ ] Filter in complex view, then download — verify only filtered items exported
- [ ] Run Daily Disposition Processing workflow — verify all submissions fetched
- [ ] All existing tests pass (31/31 ConceptMapping, 1204 total)

### DIFF
--- a/__tests__/lib/prolific-api.test.ts
+++ b/__tests__/lib/prolific-api.test.ts
@@ -318,6 +318,10 @@ describe('Prolific API Client', () => {
       expect(result.results[0].id).toBe('sub-1')
       expect(result.results[1].id).toBe('sub-2')
       expect(global.fetch).toHaveBeenCalledTimes(2)
+      // Verify second call used the correct paginated path
+      const secondCall = (global.fetch as jest.Mock).mock.calls[1][0]
+      expect(secondCall).toContain('/submissions/')
+      expect(secondCall).toContain('page=2')
     })
   })
 
--- a/src/components/tabs/concept-mapping/download-buttons.tsx
+++ b/src/components/tabs/concept-mapping/download-buttons.tsx
@@ -52,7 +52,16 @@ async function exportExcel(headers: string[], data: RowData[], filename: string)
   }))
 
   for (const row of data) {
-    ws.addRow(Object.fromEntries(headers.map((h) => [h, row[h] ?? ''])))
+    ws.addRow(
+      Object.fromEntries(
+        headers.map((h) => {
+          let val = String(row[h] ?? '')
+          // Prevent formula injection in Excel cells
+          if (/^[=+\-@\t\r]/.test(val)) val = `'${val}`
+          return [h, val]
+        })
+      )
+    )
   }
 
   ws.getRow(1).font = { bold: true }


clarkemoyer:

1. **`src/components/tabs/concept-mapping/download-buttons.tsx`**: The buttons don’t specify type="button". In React/HTML the default is type="submit", which can cause unintended form submissions if this component is ever rendered inside a <form> (now or in future reuse). Set explicit type="button" for both buttons.

2. **`src/components/tabs/concept-mapping/download-buttons.tsx`**: This new shared exporter duplicates the CSV/Excel export logic that already exists in concept-mapping/simple.tsx. Consider reusing this component (or extracting shared helpers) in the Simple view as well to avoid future drift between implementations.

3. **`src/components/tabs/concept-mapping/summary.tsx`**: Download buttons were added, but ConceptMappingSummary.test.tsx doesn’t assert their presence. Please add a test that the Excel and CSV download buttons render (similar to the Simple view tests) to prevent regressions.

4. **`src/components/tabs/concept-mapping/complex.tsx`**: Bulk download controls were added to the Complex view, but there are currently no unit tests covering ConceptMappingComplex at all. Consider adding a basic render test that verifies the download buttons appear and that filtering affects the exported row count to match the feature’s contract.

clarkemoyer:

1. **`src/components/tabs/concept-mapping/download-buttons.tsx`**: CSV export wraps values in quotes but does not mitigate CSV/Excel formula injection (cells starting with '=', '+', '-', '@'). If any exported fields could ever include user-controlled text, opening the CSV in Excel can execute formulas. Consider sanitizing each cell for CSV by prefixing a leading apostrophe or tab when the first character matches those formula prefixes (while keeping the current quote escaping).

2. **`src/lib/prolific-api.ts`**: `fetchAllPages` assumes `meta.next` is always an absolute URL and calls `new URL(nextUrl)`. If the API ever returns a relative next link (common in some paginated APIs), this will throw and break pagination. Consider constructing the URL with a base (e.g., `new URL(nextUrl, PROLIFIC_API_BASE_URL)`) and/or handling already-path-like values without parsing.
```suggestion
    // meta.next may be an absolute or relative URL — normalize to an endpoint path
    let path: string
    try {
      const url = new URL(nextUrl, PROLIFIC_API_BASE_URL)
      path = url.pathname.replace(/^\/api\/v1/, '') + url.search
    } catch {
      // Fallback: treat nextUrl as already being a path/query string
      path = nextUrl
    }

    // Ensure the endpoint path starts with a leading slash for makeApiRequest
    if (!path.startsWith('/')) {
      path = `/${path}`
    }
```

3. **`src/components/tabs/concept-mapping/complex.tsx`**: The complex concept-mapping view now has new download behavior (bulk Excel/CSV export that should respect the active search filter), but there are no unit tests covering the complex view at all. Since Simple and Summary views have tests asserting the download buttons render, add at least a basic render test for the complex download buttons (and ideally a test that filtering changes the exported row set).

4. **`src/components/tabs/concept-mapping/summary.tsx`**: The PR/issue linkage looks inconsistent: the PR description says this completes issue #574, but the referenced issue text is about fixing column widths on concept mapping pages. This change adds download buttons (and a Prolific pagination fix) but doesn’t address column width behavior; consider updating the PR description/linked issue(s) so the scope and tracking match.

clarkemoyer:

1. **`src/components/tabs/concept-mapping/download-buttons.tsx`**: The CSV formula-injection mitigation only checks the first character; values starting with whitespace before a dangerous character (e.g. " =HYPERLINK(...)" or "\n=...") can still be interpreted as formulas by spreadsheet apps. Consider checking for dangerous characters after leading whitespace (e.g., `^\s*[=+\-@]`) and/or normalizing leading whitespace before applying the guard.
```suggestion
        // Prevent CSV formula injection: prefix dangerous chars (after any leading whitespace) with a single quote
        if (/^\s*[=+\-@]/.test(val)) val = `'${val}`
```

2. **`src/lib/prolific-api.ts`**: `fetchAllPages` normalizes absolute `meta.next` URLs by stripping `/api/v1`, but for relative `meta.next` values it uses the string as-is. If Prolific returns relative links like `/api/v1/...`, the next request will become `${PROLIFIC_API_BASE_URL}/api/v1/...` (duplicated `/api/v1`) and fail. Consider applying the same normalization to relative URLs (e.g., strip a leading `/api/v1` when present) before passing to `makeApiRequest`.
```suggestion
      // Relative URL — normalize by stripping leading /api/v1 if present
      path = nextUrl.replace(/^\/api\/v1/, '')
```

clarkemoyer:

1. **`src/components/tabs/concept-mapping/summary.tsx`**: The PR description says it “completes issue #574 scope”, but the referenced issue text/image is about fixing column widths, while this PR mainly adds download buttons and Prolific pagination. Please update the PR description / issue reference so the linked issue matches the changes being shipped (or link the correct issue for downloads/pagination).

PR #776, opened 2026-04-05
fix: prevent false incident issues when 0 smoke checks fail

clarkemoyer:
## Summary\n\n- Fix false incident issues created when early workflow steps (build, curl, npm ci) fail but no actual smoke test checks failed\n- The `if: failure()` condition fires on ANY step failure, but `ROUTE_FAIL`/`ASSET_FAIL` env vars are never set when early steps fail, so `totalFail` computes to 0 — creating misleading \"0 check(s) failed\" issues\n\n## Changes\n\n- Add `id: verdict` to the \"Fail if any checks failed\" step\n- Gate incident issue creation on `steps.verdict.outcome == 'failure'` instead of job-level `failure()`\n- Gate commit status steps on `steps.verdict.outcome` for consistency\n- Add `totalFail === 0` guard in incident issue script as belt-and-suspenders safety\n\n## Test plan\n\n- [ ] Verify no false incident issues created on next deployment\n- [ ] Verify real failures still create incident issues (trigger with workflow_dispatch against a bad URL)\n\nhttps://claude.ai/code/session_01Y3snmesJFKAyvL7VkaKZYZ

### DIFF
--- a/.github/workflows/post-deploy-smoke.yml
+++ b/.github/workflows/post-deploy-smoke.yml
@@ -694,7 +694,7 @@ jobs:
       # ── Set commit status on failure ──────────────────────────────────
 
       - name: Set commit status (failure)
-        if: always() && (steps.verdict.outcome == 'failure' || (failure() && steps.verdict.outcome != 'success'))
+        if: always() && (steps.verdict.outcome == 'failure' || failure())
         uses: actions/github-script@v8
         with:
           script: |


clarkemoyer:

1. **`.github/workflows/post-deploy-smoke.yml`**: The commit status steps are now gated only by `steps.verdict.outcome`. If an earlier step (e.g., `npm ci` / `npm run build`) fails before any smoke-check env vars are set, the `verdict` step will compute `TOTAL_FAIL=0` and succeed, causing the workflow to set a **success** commit status even though the job actually failed. Consider additionally gating the success status on overall job success (e.g., `success()`/`job.status == 'success'`), and emitting a distinct `error`/`failure` status (or skipping statuses) when the job fails for non-smoke-check reasons.

clarkemoyer:
**Copilot Review Cycle:** ✅ Passed after 2 round(s) — review clean, all CI checks green. Ready to merge.

PR #1371, opened 2026-04-08
fix: full-site search — CI, ARIA, XSS, basePath nav, debounce, generator sync

clarkemoyer:
`SearchInput` was never mounted in the header, breaking all E2E tests. Multiple ARIA violations, an XSS vector in `dangerouslySetInnerHTML`, `window.location.href` breaking basePath deployments, and stale-state on Escape were also present.

### Header / CI (`src/components/header/index.tsx`)
- Mounted `SearchInput` inline in the desktop nav; removed unused `isSearchOpen`/`searchQuery`/`LiaSearchSolid`
- Removed duplicate `aria-label` prop (`react/jsx-no-duplicate-props` lint error)

### ARIA (`search-input.tsx`)
- `role="combobox"` on `<input>` — `aria-expanded` is invalid on `searchbox`; combobox is the correct pattern
- `aria-expanded` reflects actual open state: `isOpen && query.trim().length > 0` (was `results.length > 0`)
- `aria-label="Search site"` added (no accessible name before)
- `role="listbox"` + `id="search-listbox"` moved to `<ul>` only when options are present; `aria-controls` updated to match; loading/no-results use `role="status" aria-live="polite"` (empty listbox is invalid ARIA)
- Stable `id="search-option-{index}"` on each `<li>` + `aria-activedescendant` synced to `selectedIndex`

### Navigation (`search-input.tsx`)
- `window.location.href` → `router.push()` on both Enter and click — respects `basePath`, avoids full reload
- `closeSearch()` helper resets `isOpen`/`query`/`results`/`selectedIndex` after navigation so the dropdown never persists on the destination page

### Debounce / stale state (`search-input.tsx`)
- 150ms debounce via `timeoutRef`; `searchRequestIdRef` invalidates in-flight responses
- Escape clears pending timeout and increments request ID; `isLoading` reset only when a timeout was pending
- `useEffect([index])` re-triggers search once index loads if a non-empty query is already present

### XSS (`src/lib/search.ts`)
- `highlightTerms()` HTML-escapes text before inserting `<mark>` tags; `&` replaced first to prevent double-escaping:

```ts
function escapeHtml(str: string): string {
  return str
    .replace(/&/g, '&amp;')
    .replace(/</g, '&lt;')
    .replace(/>/g, '&gt;')
    .replace(/"/g, '&quot;')
    .replace(/'/g, '&#39;')
}
```

- `createSnippet()` moved after `.filter().sort().slice()` — only computed for results that render

### Index generator / `public/search-index.json`
- Generator imports `faqs` directly (typed) instead of regex-parsing the source file
- FAQ content now indexes `question + " " + answer`; `category` field added to all main pages
- Committed `search-index.json` regenerated: 4 FAQs with full answer text, categories on all entries

### Docs / tests
- `SEARCH_IMPLEMENTATION.md`: removed contradictory "URLs are already prefixed in the index" claim; clarified index is generated via manual script, not `next build`; `aria-controls` section updated to reference `#search-listbox`
- `__tests__/lib/search.test.ts`: 3 XSS regression tests for `highlightTerms` (`<script>` injection, `&` double-escape, highlight+escape combo); test name corrected to match actual assertion
- E2E: no-results query changed to `kvjwxfbqps` (no common substring matches); flaky immediate-listbox assert replaced with `await firstResult.toBeVisible()`; navigation assertion via `waitForURL`; `openHeaderSearch` fallback removed in favour of explicit desktop viewport


clarkemoyer:

1. `src/components/search/search-input.tsx`: The search input has no accessible name (no <label>, aria-label, or aria-labelledby). Placeholders are not a reliable accessible name, and this will be read poorly by assistive tech and can be flagged by automated a11y checks. Add an explicit label (visually hidden is fine) or set aria-label on the input and ensure it stays in sync with the surrounding "Site search" landmark label.
2. `src/components/search/search-input.tsx`: The dropdown container uses role="listbox" even when rendering the loading state, but the loading markup contains no children with role="option". This is invalid ARIA structure (listbox requires option/group children) and can trigger screen reader/axe issues. Consider rendering a separate non-listbox container for loading (e.g., role="status") and only applying role="listbox" when options are present.
3. `src/components/search/search-input.tsx`: Similarly, when there are no results the element with role="listbox" renders a plain <div> message instead of role="option" children. If you keep role="listbox", render an aria-disabled option for the empty state or switch the container role for the empty-state message to avoid invalid ARIA required-children relationships.
4. `src/components/search/search-input.tsx`: Using dangerouslySetInnerHTML with highlightTerms(result.snippet, query) can introduce XSS if search-index.json ever contains HTML (e.g., if future index generation pulls from page markup). Since highlightTerms injects <mark> tags without escaping, any '<'/'>' in snippet content would be interpreted as HTML. Prefer rendering highlights without innerHTML (split into text/mark nodes) or escape snippet content before inserting markup.
5. `scripts/generate-search-index.ts`: The direct-execution guard `if (require.main === module)` can throw a ReferenceError if this file is ever executed/loaded in an ESM context (tsx can run in ESM mode). Other scripts in this repo use an ESM-safe guard (`typeof require !== 'undefined' && require.main === module`). Consider matching that pattern here to avoid brittle runtime failures.
```suggestion
if (typeof require !== 'undefined' && require.main === module) {
```

clarkemoyer:

1. `src/components/search/search-input.tsx`: Selecting a result navigates via `router.push(...)` but does not close/reset the search UI. Because the header persists across route changes in the App Router, the dropdown can remain open on the destination page. Consider closing the popup and clearing query/results/selection when a result is chosen (both click and Enter).
2. `tests/full-site-search.spec.ts`: This test asserts `[role="listbox"]` is visible immediately after `fill()`, but the component shows a debounced loading state first and only renders the listbox `<ul>` once results are available. This is likely to be flaky; consider asserting `#search-results` becomes visible, or waiting for the first `[role="option"]` instead.
```suggestion
    // Results should appear once the debounced search has populated options
    const firstResult = page.locator('[role="option"]').first()
    await expect(firstResult).toBeVisible()
```
3. `scripts/generate-search-index.ts`: The generator currently indexes FAQ `content` using `faq.question` only. That makes FAQ results much less useful (and contradicts the committed `public/search-index.json`, which includes answer text). Consider indexing `faq.answer` (or `question + " " + answer`) so search/snippets can match what users actually read.
```suggestion
    const faqContent = `${faq.question} ${faq.answer}`

    items.push({
      id: `faq-${id++}`,
      url: '/barriers#faqs',
      title: faq.question,
      description: 'Frequently asked question',
      content: truncateContent(faqContent),
```
4. `public/search-index.json`: This committed `search-index.json` appears out of sync with `src/data/faqs.ts` and the generator: `faqs.ts` exports 4 FAQs, but only 3 FAQ entries are present here. If this file is intended to be generated+committed, it should be regenerated so it includes all FAQs and matches the generator’s schema/content.
```suggestion
    "category": "FAQ"
  },
  {
    "id": "faq-4",
    "url": "/barriers#faqs",
    "title": "How can organizations measure successful technology adoption?",
    "description": "Frequently asked question",
    "content": "Organizations can measure adoption success through usage rates, employee proficiency, process improvements, stakeholder satisfaction, and progress toward intended business outcomes.",
    "category": "FAQ"
```
5. `SEARCH_IMPLEMENTATION.md`: The basePath section says “All URLs include basePath automatically” and also “URLs are already prefixed in the index,” which is contradictory and can lead to double-prefixing if someone follows it literally. Since navigation uses `router.push(document.url)`, the index URLs should generally be basePath-less (Next will apply basePath), and the docs should reflect that.
```suggestion
- Indexed document URLs should remain basePath-less; Next.js applies `basePath` during navigation
- Works on both custom domain and GitHub Pages deployment

## GitHub Pages Basepath Handling

The component automatically handles the `NEXT_PUBLIC_BASE_PATH` environment variable for loading the search index, while document URLs in the index remain relative to the app root:

```tsx
// In search.ts
const basePath = process.env.NEXT_PUBLIC_BASE_PATH || ''
const response = await fetch(`${basePath}/search-index.json`)

// Store document.url values without the basePath prefix.
// Navigation (for example, router.push(document.url)) should let Next.js apply basePath.
```
6. `__tests__/lib/search.test.ts`: `highlightTerms()` now HTML-escapes before inserting `<mark>` tags, but the test suite doesn’t assert the escaping behavior (e.g., `<script>` / ``). Adding a regression test for escaping would better lock in the XSS fix.

clarkemoyer:
@copilot ok check for new merges to main and then consider a new review round

clarkemoyer:
@copilot ok is this ready to merge?

clarkemoyer:
@copilot ok are we ready to merge. If so lets merge and then watch all the way to post deployment smoke tetsing for this merge

clarkemoyer:
@copilot how we looking?

PR #1373, opened 2026-04-08
fix: pin pyzotero[mcp] to v1.11.0 in CLAUDE.md and .vscode/mcp.json.example

clarkemoyer:
Unpinned `uvx --from "pyzotero[mcp]"` in both the CLAUDE.md setup command and the VS Code MCP example config could silently install a newer release and break local Zotero MCP setups.

## Changes

- **`CLAUDE.md`**: Pin `uvx --from` spec to `pyzotero[mcp]==1.11.0`, matching the version already stated in the package description line
- **`.vscode/mcp.json.example`**: Pin `--from` arg to `pyzotero[mcp]==1.11.0` for the same reason

```jsonc
// .vscode/mcp.json.example — before
"args": ["--from", "pyzotero[mcp]", "pyzotero-mcp"]

// after
"args": ["--from", "pyzotero[mcp]==1.11.0", "pyzotero-mcp"]
```


clarkemoyer:

1. `src/app/making-of-tabs/integrations/zotero/page.tsx`: Breadcrumb styling/labels here differ from the established pattern used by the other integration deep-dive pages (they use the single-character separator `›` and label the hub link as “Technical Integrations”). To keep the integrations section consistent, update this breadcrumb to match (see e.g. `src/app/making-of-tabs/integrations/cloudflare/page.tsx:19-38`).
2. `src/app/making-of-tabs/integrations/page.tsx`: The page intro (above) still says “three major platform integrations” and only lists Qualtrics/Prolific/Google Analytics, but this PR adds Zotero as an integration deep-dive and section. Please update the intro copy to reflect the current set of major integrations (count + list) so the overview matches the rest of the page.
3. `src/app/making-of-tabs/integrations/page.tsx`: This environments table now lists `zotero-prod` as “1 workflow”, but the repo currently has at least two workflows using `environment: zotero-prod` (e.g. `zotero-api-smoke.yml` and `zotero-create-collections.yml`). Please update the workflow count (or adjust the wording if you’re only counting a subset) to avoid misleading ops documentation.
```suggestion
                    <td className="py-2">2 workflows</td>
```
4. `CLAUDE.md`: `zotero-prod` is documented here as “✅ Active (1 workflow)”, but the repo has at least two workflows targeting that environment (`.github/workflows/zotero-api-smoke.yml` and `.github/workflows/zotero-create-collections.yml`). Update the count/status so the environment summary stays accurate.
```suggestion
| `zotero-prod`    | Zotero Web API v3       | 1 secret, 3 vars  | ✅ Active (2 workflows)    |
```
5. `CLAUDE.md`: In this snippet, `api_key` is referenced but never defined, which makes the cloud API example non-runnable/copy-pasteable. Use a concrete source for the API key (e.g., from an environment variable) or rename it to match the documented `ZOTERO_API_KEY` variable.
6. `CLAUDE.md`: This note says “VS Code will prompt for OAuth tokens when connecting”, but Zotero MCP is configured as a local `stdio` server and doesn’t use OAuth; only the Qualtrics server here requires an OAuth token input. Clarify this line so Zotero setup isn’t confusing (e.g., “Qualtrics will prompt for OAuth…”).
```suggestion
# VS Code will prompt for OAuth tokens when connecting to Qualtrics/GitHub MCPs
```

PR #1596, opened 2026-04-14
Add References sections to 10 Branch 1 bibliography pages

clarkemoyer:
## Summary
Adds missing References sections to the 10 Branch 1 bibliography pages that lacked them, bringing the total from 32/42 to 42/42 pages with References.

| Page | Model | Refs |
|---|---|---|
| 1-1 | Theory of Reasoned Action (TRA) | 13 |
| 1-4 | Model of Innovation Resistance | 10 |
| 1-5 | Status Quo Bias | 10 |
| 1-6 | Technology Acceptance Model (TAM) | 5 |
| 1-7 | Theory of Planned Behavior (TPB) | 5 |
| 1-9 | Intrinsic/Extrinsic Motivation | 6 |
| 1-11 | Task-Technology Fit (TTF) | 10 |
| 1-17 | Value-Based Adoption Model (VAM) | 12 |
| 1-18 | TRAM - Lin, Shih & Sher | 10 |
| 1-21 | Technology Readiness Index 2.0 (TRI 2.0) | 10 |

All reference text is verbatim from the CRP Lit Review Part 2 source document. No en/em dashes.

Closes #1548

## Test plan
- [x] `npm run build` succeeds
- [x] `npm run lint` - 0 errors
- [x] `npm run format` - clean
- [ ] Visual: all 10 pages render References as ordered list under H2 heading
- [ ] Verify 42/42 bibliography pages now have References markers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1553, opened 2026-04-14
Bibliography Series Standardization: canonical 16-section template + full R1-R3+grumpy review across 42 pages


clarkemoyer:
## Goal

Bring every bibliography page in the TABS Bibliography Series (`src/app/bibliography-1-*` and `src/app/bibliography-2-*`, 42 pages total) to a **single canonical 16-section template**, with every page having completed a **full R1-R3+grumpy review cycle** against its source publication.

The website is the canonical representation; standardizing the website also standardizes CRP Appendix G content.

## Scope Evolution

| Scope | Status | Count | % |
|-------|--------|-------|---|
| **Original 14-section canonical** (this umbrella's initial scope) | 41 / 42 pages | 41 | 98% |
| **Expanded 16-section canonical** (session-expanded: adds "What Does the Model Measure?" Section 6 + splits "Further Reading" as Section 15) | 24 / 42 pages | 24 | 57% |
| **Full R1-R3+grumpy review cycle** (session-introduced: structural + softening + refs/rebase + pre-merge audit) | 24 / 42 pages | 24 | 57% |
| **Current 16-section + full-review standard** | 24 / 42 pages | 24 | 57% |

## Canonical 16-Section Template (current)

1. Model Identification (or Framework Identification)
2. Theory Publication Information
3. Citation Information
4. Why Was the Model Created?
5. Core Concepts and Definitions
6. **What Does the Model Measure?** *(added during expansion - codifies measurement-model vs prescriptive-framework distinction)*
7. Preceding Models or Theories
8. Describe The Model
9. Key Contributions
10. Internal Validity
11. External Validity
12. Relevance to Technology Adoption
13. Following Models or Theories
14. References *(body-cited only)*
15. **Further Reading** *(added during expansion - split from References so non-cited suggestions are separate)*
16. Series Navigation

## Review Methodology

Each page must pass four phases before its child issue is fully complete:

- **R1 Structural audit:** canonical 16-section template applied; factual corrections; anachronism detection; source-availability flagging ("Option B") for PDF-unavailable books
- **R2 Softening pass:** hedge "Demonstrated/Established/Spawned" language; add vendor-framework source flags; add sample-scope notes on empirical papers
- **R3 References + rebase:** merge main to drop orphan commits; verify every in-body `#ref-*` cite resolves to matching `<li id="ref-*">` entry; single-file PR diff
[...]

Issue #1539, opened 2026-04-14
Site style guide, CI typography guard, Vale + Microsoft, and Zotero-backed APA citation validation (dry-run first)


clarkemoyer:
## Summary

Establish a single authoritative **site style guide** (`STYLE.md`) and a mechanical **CI enforcement chain** so every public-facing change is checked before merge. The chain combines three public toolkits plus one custom checker, all runnable locally and in CI:

1. **grep-based typography guard** (forbidden Unicode, HTML entities, punctuation-on-newline)
2. **Vale** with the Microsoft Writing Style Guide package (clarity, wordiness, contractions, acronyms, sentence-case headings, etc.)
3. **APA 7 citation validator** backed by the **Zotero "Website Citations" collection** as the single source of truth for references
4. **Custom `tabs-style-check.py`** for project-specific rules (construct names, statistical notation, PII patterns) not covered above

Every rule in `STYLE.md` is mechanically verifiable. No human-review rules; no rules without a `Check:` command.

**Dry-run first.** Every new workflow lands in `report-only` mode with **Article 1 (`src/app/article-1-*/`) as the initial scope**. We observe output for at least one week, tune false positives, expand scope, and only then flip to `fail-on-error`. Rationale: we have an established pattern of agent PRs over-correcting (PR #1487 mojibake -> en-dash regression, PR #1528 partial revert history), so the risk of a hot-turned-on CI gate breaking in-flight work is higher than the risk of a slow rollout.

This issue is the prerequisite to the site-wide remediation tracked in **#1538**. Land this first in dry-run. Remediate #1538. Promote to enforce.

## What already exists (do not duplicate)

Before writing any new code, the implementing agent must read and build on:

- **`zotero-prod` GitHub environment** - already provisioned with `ZOTERO_API_KEY` (secret), `ZOTERO_USER_ID` / `ZOTERO_BASE_URL` (vars). Some newer PRs reference additional vars `ZOTERO_LIBRARY_ID`, `ZOTERO_LIBRARY_TYPE`, `ZOTERO_COLLECTION_NAME` - add these if they are not yet present
- **`.github/workflows/zotero-api-smoke.yml`** (merged via #1283) - manual read-only smoke test
- **`.github/workflows/zotero-create-collections.yml`** - creates the parent "Website Citations" collection plus 6 subcollections (Bibliography Branch 1, Branch 2, Articles, Concept Maps, Results & Analysis, Software & Tools)
[...]

Issue #1344, opened 2026-04-07
feat: standardize citation format across all website pages


clarkemoyer:
## Problem

The TABS website uses **inconsistent citation methods** across different page types:

| Page Type | Citation Method | Example |
|-----------|----------------|---------|
| Bibliography pages (43) | DOI links + structured metadata | `href="https://doi.org/10.2307/249008"` |
| Branch 1 articles (1-1 through 1-7) | Footnotes `[1]` + DOI-linked references section | References with `doi.org` hrefs |
| Branch 2 articles (2-1 through 2-7) | Inline author mentions + APA-style references (no DOI links) | Plain text: `Barney, J. B. (1991)` |
| Concept maps | JSON data with `Theoretical Grounding (Source)` field | `"Hofstede (1980); Cameron & Quinn (2006)"` |
| Article introductions | Mix of both styles | Some DOI links, some plain text |

This inconsistency means:
1. **Automated citation validation** (#1298) cannot reliably extract all citations
2. **18 article references** couldn't be matched to Zotero because they lack DOIs
3. Branch 2 articles have zero DOI links despite citing well-known papers with DOIs
4. Some references use first author only (e.g., "Pisano, G., & Shuen, A. (1997)") when the canonical entry is under a different first author (Teece)

## Proposed Standard

Adopt a **consistent citation approach** across all page types:
1. Every reference should have a `doi.org` link where a DOI exists
2. References sections should use a consistent format (APA 7th with DOI links)
3. Consider a shared `<Citation doi="10.xxx" />` component that auto-renders from Zotero data
4. Concept map data should include DOI fields alongside author-year text

## Scope

- [ ] Audit all 16 article pages for reference format consistency
- [ ] Add DOI links to Branch 2 article reference sections (currently plain text)
- [ ] Standardize reference format across Branch 1 and Branch 2
- [ ] Update concept map JSON data with DOI fields
- [ ] Consider a reusable Citation component for future consistency

## 18 References Not Matched to Zotero

These article references couldn't be automatically matched — most are likely in Zotero under different author names or formats:

- Anushka, K., & Madhan Raj, M. S. (2024)
- Bagozzi, R. P., & Warshaw, P. R. (1992) — likely Davis, Bagozzi, & Warshaw
- Chan, H. C., & Gupta, S. (2007) — likely Kim, Chan, & Gupta
- Chrissis, M. B., & Weber, C. V. (1993) — CMMI book
[...]

PR #1597, opened 2026-04-14
Zotero-backed bibliography reference validation (CI + script)

clarkemoyer:
## Summary

Adds automated validation that cross-checks every reference in the 42 bibliography pages against the Zotero library. This is the enforcement mechanism for the APA 7th citation standard established in #1553.

### What it does
- **Python script** (`scripts/validate-bibliography-references.py`): Extracts author-year pairs from each page's References `<ol>`, queries Zotero API, reports unmatched
- **CI workflow**: Triggers on any PR touching `bibliography-*/page.tsx`, posts a PR comment with match/unmatch table
- **Non-blocking** for now (warns, doesn't fail) - becomes blocking once Zotero coverage reaches 90%

### Current baseline
- 242 references across 32 pages (10 pages still missing References - #1596 adds them)
- 34% matched in Zotero (83/242)
- 159 references need to be added to Zotero as the standardization work (#1553) progresses

### How agents should use this
When working any of the 30 bibliography standardization issues under #1553:
1. Add/verify references in APA 7th format
2. Ensure each cited work exists in Zotero (tagged `website-citation`)
3. The CI check will report gaps on the PR automatically

### Related issues
- #1298 - CI workflow to validate website citations against Zotero
- #1299 - Validate bibliography pages match Zotero entries  
- #1553 - Bibliography standardization umbrella (APA 7th + Zotero requirement)
- #1344 - Standardize citation format across all pages
- #1296 - Zotero integration - website citation management
- #1539 - Site style guide + Zotero-backed APA validation

Partial implementation of #1298 and #1299.

## Test plan
- [x] Script runs locally with `--local` flag against Zotero desktop
- [x] Script extracts references from all 32 pages with References sections
- [x] JSON report generated with match/unmatch details
- [x] `npm run build` passes (no website changes)
- [ ] CI workflow triggers on bibliography PR and posts comment


clarkemoyer:
**Copilot Review Cycle (Round 1/14):** Timed out waiting for Copilot review. You can re-trigger manually.

clarkemoyer:

|---|---|---|---|---|
| 1 | 0 | — | ✅ Clean | 2026-04-14 04:39 |

### Result

✅ All automated review comments addressed in 2 round(s). Ready for final human review.


**Copilot Review Cycle:** ✅ Passed after 2 round(s) - review clean, all CI checks green. Ready to merge.

clarkemoyer:

1. `scripts/validate-bibliography-references.py`: The bibliography pages render references as `<li id="ref-...">…</li>`, so this regex only matching `<li>…</li>` will fail to extract any references. This will make the validator report 0 references (and can cascade into later errors). Consider matching `<li[^>]*>` (and ideally keying off `id="ref-"`) or using a small HTML/JSX parser instead of a strict `<li>` literal match.

2. `scripts/validate-bibliography-references.py`: `total_refs` can be 0 (e.g., if extraction fails), which will raise a ZeroDivisionError here. Guard the percentage calculation (e.g., print 0% when total_refs is 0) so the script can still write a report for CI.

3. `scripts/validate-bibliography-references.py`: This only requests up to 100 items per collection and doesn’t paginate, so the index can be incomplete for larger collections. That will produce false “NOT IN ZOTERO” warnings; consider using pyzotero pagination helpers (e.g., `zot.everything(...)`) or looping with `start`/`limit` until exhaustion.

4. `.github/workflows/validate-bibliography-zotero.yml`: The job-level permissions are likely insufficient: actions/checkout requires `contents: read`, and posting/updating PR issue comments via `github.rest.issues.*` requires `issues: write`. With only `pull-requests: write`, the checkout and/or comment step may fail. Add `contents: read` and `issues: write` (and keep `pull-requests: write` only if you also need PR-specific APIs).
5. `.github/workflows/validate-bibliography-zotero.yml`: `python ... || true` swallows all failures (including real script errors and missing secrets), but the next step unconditionally reads `bibliography-validation-report.json`. If the script exits before writing the report, the PR comment step will throw and fail the workflow. Prefer `continue-on-error` only for the “unmatched refs” exit code, and/or ensure the script always writes a report (even on error) and have the GitHub Script step handle a missing report gracefully.
6. `.github/workflows/validate-bibliography-zotero.yml`: This workflow uses `actions/checkout@v4`, but the repo’s workflows consistently use `actions/checkout@v6` (Node 24–compatible). Aligning to v6 helps avoid runtime/deprecation issues and keeps action versions consistent across CI.

7. `.github/workflows/validate-bibliography-zotero.yml`: This uses `actions/github-script@v7`, while other workflows in the repo use `actions/github-script@v9`. Upgrading keeps the workflow consistent and reduces risk under the forced Node 24 runtime.

8. `scripts/validate-bibliography-references.py`: The docstring says the script “optionally checks that matched items are in the correct collection”, but the current implementation only indexes items and reports unmatched references. Either implement that collection-membership check or adjust the docstring so CI users aren’t expecting functionality that isn’t there.


clarkemoyer:

1. `scripts/validate-bibliography-references.py`: `zot.collection_items(..., itemType="-attachment || note")` does not look like a valid Zotero API `itemType` filter (the API expects a single itemType, with optional negation like `-attachment`). As written, this likely returns zero items and makes the index empty, causing every reference to appear unmatched. Use supported filtering (or filter attachments/notes client-side) so the index is built from the intended items.

2. `scripts/validate-bibliography-references.py`: Zotero creator objects can be single-field (e.g., corporate authors) using `name` instead of `firstName`/`lastName`. Indexing only `creator.get("lastName")` will skip those items entirely, leading to false "NOT IN ZOTERO" results for org-authored works (e.g., DoD/Open Group). Consider falling back to `creator.get("name")` when `lastName` is missing, and normalize consistently with the extracted page author string.
3. `.github/workflows/validate-bibliography-zotero.yml`: The `continue-on-error: true` on the validation step makes the workflow green for both expected "unmatched" exits (code 1) and real script failures (code 2, API outage, auth misconfig). Since the script already differentiates exit codes, consider allowing exit code 1 but failing the job on exit code 2 so CI still catches broken validation runs.

4. `.github/workflows/validate-bibliography-zotero.yml`: If the script writes an error report (it includes an `error` field), the PR comment currently ignores it and will misleadingly show 0 matches/unmatched. Consider surfacing `report.error` prominently in the comment body (and possibly skipping percentage calculations) when the report indicates an error, so failures are actionable.

5. `.github/workflows/validate-bibliography-zotero.yml`: On PRs from forks, GitHub Actions will not provide `secrets`/`vars` from the `zotero-prod` environment and the `GITHUB_TOKEN` may be read-only, so this job will at best produce a "missing env" error report and may fail when trying to post a comment. Consider guarding the validation/comment steps to run only when `github.event.pull_request.head.repo.fork == false` (or when `github.repository == github.event.pull_request.head.repo.full_name`) to avoid noisy failures on forked PRs.
6. `scripts/validate-bibliography-references.py`: `validate_page()` takes a `filepath` argument but the caller passes `slug`, and the parameter is unused inside the function. Consider removing the unused parameter or renaming it to reflect the actual value passed, to avoid confusion and make type intent clearer.
7. `scripts/validate-bibliography-references.py`: `Optional` is imported from `typing` but never used. Removing unused imports helps keep the script clean and avoids lint noise if Python linting is added later.

8. `scripts/validate-bibliography-references.py`: `status` is computed but never used. Consider removing it (or using it in output/report) to keep the validation loop focused and avoid dead code.


clarkemoyer:

1. `scripts/validate-bibliography-references.py`: `build_zotero_index()` is annotated as returning `dict`, but it actually returns a 2-tuple `(index, item_count)`. This makes the function signature incorrect for type checkers/IDEs and can hide real bugs. Update the return type annotation (and ideally the `index` key/value types) to match the actual tuple being returned.
2. `scripts/validate-bibliography-references.py`: The Zotero collection keys are hard-coded in the script. If collections are recreated (or this is run against a different Zotero library), the keys will change and validation will silently degrade (fewer items indexed → more false “NOT IN ZOTERO”). Consider making these keys configurable via environment variables (or looking collections up by name once) and include a clear error when none of the configured collections can be fetched.
3. `.github/workflows/validate-bibliography-zotero.yml`: The workflow only triggers on changes to `src/app/bibliography-*/page.tsx`. If someone updates the validator script or this workflow file itself, the check won’t run on that PR, which makes it easy to merge a broken validator. Consider expanding the `paths:` filter to also include `scripts/validate-bibliography-references.py` and `.github/workflows/validate-bibliography-zotero.yml` (and any shared Zotero config files it depends on).

4. `scripts/validate-bibliography-references.py`: The docstring’s Usage/Environment section doesn’t mention supported flags (`--dry-run`, `--verbose/-v`) or `VALIDATION_REPORT_PATH`, even though the script implements them. Updating the usage docs will help contributors run the script correctly and understand how CI controls the report output location.


clarkemoyer:

1. `scripts/validate-bibliography-references.py`: `first_author` extraction keeps trailing punctuation (e.g., corporate authors in bibliography pages are written as "Microsoft." / "The Open Group."), but the Zotero creator `name` field is typically stored without that trailing period. This will cause systematic false negatives when matching `(author, year)` pairs. Consider normalizing both extracted authors and Zotero creator names (trim trailing punctuation/whitespace, and possibly collapse internal whitespace) before building/looking up keys.
2. `scripts/validate-bibliography-references.py`: Cloud-mode Zotero access always uses pyzotero's default base URL. The repo already has a `ZOTERO_BASE_URL` variable used by `.github/workflows/zotero-api-smoke.yml`; it would be more robust to honor an env override (e.g., `ZOTERO_BASE_URL`) and set the client endpoint accordingly, so CI and local runs stay consistent if the base URL ever changes.
3. `.github/workflows/validate-bibliography-zotero.yml`: The workflow posts PR comments via the Issues API (`issues.*`) and does not use any Pull Requests API calls. To follow least-privilege, `pull-requests: write` can likely be removed (keep `issues: write` and `contents: read`).

4. `.github/workflows/validate-bibliography-zotero.yml`: Header comment says the workflow is "Non-blocking (warns but doesn't fail CI)", but the run step explicitly fails the job for exit codes >=2 (e.g., Zotero outage / auth misconfig). Either clarify the comment to specify it's non-blocking only for unmatched references (exit code 1), or make the workflow fully non-blocking as advertised.


clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The workflow sets `ZOTERO_API_KEY`/`ZOTERO_USER_ID` but never passes `ZOTERO_BASE_URL`. Other Zotero workflows in this repo rely on `vars.ZOTERO_BASE_URL`, and the script supports a base URL override; without wiring it through here, the validation may silently hit the wrong endpoint (or fail if the environment is configured to use a non-default base URL). Consider exporting `ZOTERO_BASE_URL: ${{ vars.ZOTERO_BASE_URL }}` for consistency with existing Zotero workflows.

2. `.github/workflows/validate-bibliography-zotero.yml`: The script goes out of its way to write a minimal JSON error report (via `_write_error_report`) for downstream CI steps, but if the script exits with a real error (exit code 2+), this job will stop before the artifact upload / PR comment steps run. If the intent is to surface actionable failure details, consider running the upload + comment steps with `if: always()` so the error report is still attached/commented even when the job ultimately fails.
3. `scripts/validate-bibliography-references.py`: The fallback “partial match” logic treats any Zotero item with the same year and a 4-character prefix overlap as a match. This can create false positives (missing citations won’t be reported) when different authors share the same first 4 letters, undermining the script’s purpose as an enforcement signal. Consider either requiring exact (author, year) matches, or reporting fuzzy matches separately (e.g., as a "possible match" suggestion) instead of treating them as validated.

4. `scripts/validate-bibliography-references.py`: When writing the JSON report, the file is opened without an explicit encoding. Since this report may include non-ASCII characters from references, consider opening the file with `encoding="utf-8"` to make behavior consistent across environments.

clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The workflow runs on `pull_request` and checks out the PR’s code while also injecting `ZOTERO_API_KEY` from the `zotero-prod` environment. That means a PR that modifies `scripts/validate-bibliography-references.py` (or this workflow) could exfiltrate the Zotero API key (the workflow is explicitly triggered by those paths). To mitigate, avoid providing secrets to untrusted PR code: e.g., run the Zotero query job from `pull_request_target` with the script sourced from the base branch, or keep `pull_request` runs in `--dry-run` (no secrets) and only run the secret-backed validation on `workflow_dispatch`/scheduled jobs or after merge.
2. `scripts/validate-bibliography-references.py`: The default Zotero collection keys are hard-coded. These keys are library-specific and can change if collections are recreated, which would make CI fail (and it’s not obvious to contributors how to discover/update them). Consider resolving collections by name at runtime (e.g., find “Website Citations” then its “Bibliography Series - Branch 1/2” subcollections), or sourcing the keys from GitHub environment vars so they can be updated without code changes.
3. `scripts/validate-bibliography-references.py`: `build_zotero_index()` fetches *every* item from each configured collection and then concatenates them, which can cause substantial redundant API pagination if collections overlap (e.g., parent + child collections). This increases runtime and the chance of Zotero rate limiting. Consider deduplicating by Zotero item key as you ingest results, and/or defaulting to a single authoritative collection (like “Website Citations”) instead of fetching multiple overlapping collections by default.

clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The workflow treats exit code 1 as a non-blocking “unmatched references” warning, but Python also returns exit code 1 for *unhandled exceptions*. That means a parsing/Zotero/API bug that crashes the script could be silently ignored and CI would still succeed. Consider making the workflow only allow code 1 when the script completed and produced a report (or update the script to guarantee unexpected errors exit with code 2+).

2. `.github/workflows/validate-bibliography-zotero.yml`: `git checkout pr-head -- src/app/bibliography-*/page.tsx` is unquoted, so the shell will expand the glob based on the *base branch* file list. If a PR adds/renames a bibliography page path, it may not be overlaid (and thus not validated). Quote the pathspec (or use git’s `:(glob)` pathspec) so the wildcard is evaluated by git against `pr-head`.

3. `.github/workflows/validate-bibliography-zotero.yml`: The PR-comment updater only calls `issues.listComments` once, which returns a single page of results (default 30). On PRs with >30 comments, the existing validation comment might not be found and the workflow could post duplicates. Consider using `github.paginate(...)` or requesting `per_page: 100` and paginating until the marker is found.

4. `scripts/validate-bibliography-references.py`: The script can raise unhandled exceptions (e.g., unexpected file contents, API response shape changes). In that case Python exits with status 1, which this repo’s workflow currently interprets as a non-blocking “unmatched references” warning, potentially masking real failures. Wrap the top-level `main()` call in a `try/except Exception` that exits with code 2 (and ideally writes an error report) so genuine script errors reliably fail CI.


clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The PR-head overlay only checks out bibliography files that exist in the PR, but it doesn’t remove bibliography pages that were deleted/renamed in the PR. In that case, the base-branch page.tsx will remain on disk and still be validated/commented on, producing incorrect results. Consider clearing `src/app/bibliography-*` (or computing removed paths) before the `git checkout pr-head -- ...` so the workspace matches the PR head exactly.

2. `.github/workflows/validate-bibliography-zotero.yml`: This workflow is triggered when the PR changes `scripts/validate-bibliography-references.py` or the workflow itself, but the job intentionally checks out the base branch and never overlays `scripts/` or `.github/` from the PR. That means changes to the validator/workflow in a PR will not actually be executed/validated by this run. Either remove these paths from the trigger (so only bibliography content changes trigger it), or add conditional logic to run PR-head versions for same-repo PRs while keeping the base-branch-only behavior for forks.

3. `scripts/validate-bibliography-references.py`: The PR description says this validates references across 42 bibliography pages, but the glob `src/app/bibliography-*/page.tsx` currently matches 43 pages because both `bibliography-1-8-personal-computing-acceptance-thompson-1991` and `bibliography-1-8-personal-computing-utilization-thompson-1991` exist. If the acceptance page is an orphan (per #1553), consider excluding it from validation (or otherwise reconciling the page set) so the script’s scope matches the stated intent.

clarkemoyer:

1. `scripts/validate-bibliography-references.py`: The documented exit codes say `0` means “All references validated”, but the script can exit `0` even when `possible_matches > 0` (fuzzy matches are explicitly “verify manually”). Either include “possible matches found” in the non-zero/warning exit condition, or update the docstring + downstream workflow expectations so exit code semantics match actual behavior.
2. `scripts/validate-bibliography-references.py`: `build_zotero_index()` swallows per-collection fetch failures (prints a warning and continues). That can silently produce an incomplete index and misleading “unmatched” results (false positives). Consider accumulating these errors and surfacing them in the JSON report (e.g., `error`/`warnings` field) and/or failing with exit code 2 when any configured collection can’t be fetched.
3. `scripts/validate-bibliography-references.py`: The docstring claims deduplication avoids “redundant API pagination”, but the code still paginates/fetches each configured collection independently and only deduplicates items after they’re retrieved. Update this comment to reflect what the code actually guarantees (deduped index + item count) to avoid confusion when tuning performance.

4. `scripts/validate-bibliography-references.py`: `validate_page()` takes a `slug` parameter but never uses it. Dropping the unused parameter (and its call-site argument) will simplify the API and avoid confusion about whether page-level logic is intended here.


clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: Because this workflow runs on `pull_request_target` with `zotero-prod` secrets, overlaying PR files via `git checkout pr-head -- ...` is vulnerable to a symlink attack: a fork PR could add `src/app/bibliography-*/page.tsx` as a symlink to `/proc/self/environ` (or similar) and the validator would read/post secret material. Mitigate by disabling symlinks during checkout (e.g., `git config --global core.symlinks false`) and/or explicitly failing if any symlink exists under `src/app/bibliography-*` after overlay before running the script.

2. `scripts/validate-bibliography-references.py`: `extract_references_from_tsx()` blindly opens the on-disk `page.tsx` path. In the `pull_request_target` workflow, a malicious PR can introduce a symlink at `src/app/bibliography-*/page.tsx` pointing outside the repo (e.g., `/proc/self/environ`) and cause the script to read and leak runner environment/secrets into logs/report. Add a guard that rejects symlinks and ensures `realpath` stays within the repo checkout before reading.

3. `scripts/validate-bibliography-references.py`: The Zotero index currently picks the first entry in `data['creators']` (regardless of `creatorType`). Zotero items often include editors, translators, etc.; indexing by a non-author can produce systematic false “unmatched” results even when the referenced work exists. Filter creators to `creatorType == 'author'` (and for corporate authors, still use the `name` field) and index using the first author per item.
4. `.github/workflows/validate-bibliography-zotero.yml`: The PR comment body includes `${d.ref}` and `${report.error}` derived from untrusted PR file content / exception strings without escaping. In a `pull_request_target` workflow this enables markdown/mention injection (e.g., `@org/team`, table breaks, HTML) from forks. Consider escaping `|`, backticks, newlines, and stripping or neutralizing `@` mentions before posting, or render the details as a code block to avoid GitHub parsing.

clarkemoyer:

clarkemoyer:

1. `scripts/validate-bibliography-references.py`: `find_collection_keys_by_name()` is reading the collection key from `coll['data']['key']`, but Zotero collection objects expose the key at the top level (`coll['key']`). As written, name-based resolution will never find any keys and the script will always fall back to the hard-coded `_default_keys` (and warn), which is brittle if collections are recreated. Read the key from `coll.get('key')` (or fall back to `coll.get('data', {}).get('key')` if needed).

2. `scripts/validate-bibliography-references.py`: The path-escape guard in `extract_references_from_tsx()` treats the *current working directory* as the repo root (`repo_root = os.getcwd()`). This makes the script fail if it's run from any subdirectory (e.g. `cd scripts && python validate-bibliography-references.py`), because `src/app/...` then appears “outside” `repo_root`. Consider resolving the actual repo root (e.g. via `git rev-parse --show-toplevel` or `Path(__file__).resolve().parents[...]`) or explicitly documenting/enforcing “run from repo root” with a clearer error message before doing the check.
3. `.github/workflows/validate-bibliography-zotero.yml`: This workflow uses `actions/upload-artifact@v4`, but the repo consistently uses `actions/upload-artifact@v7` elsewhere (e.g. `.github/workflows/ci.yml:72-77`, `.github/workflows/daily-pipeline.yml:60-66`). Aligning on the same major version reduces maintenance risk (Node runtime support, behavior differences) and keeps workflows consistent.

clarkemoyer:

1. `scripts/validate-bibliography-references.py`: `pages = glob.glob("src/app/bibliography-*/page.tsx")` depends on the current working directory being the repo root. If a developer runs this script from another directory (e.g., `scripts/`), it will find 0 pages and silently produce a misleading “0 references” success report. Consider basing the glob on `_find_repo_root()` (or `os.chdir(repo_root)` once) and treating `len(pages) == 0` as an error exit (code 2) so CI/local runs can’t false-green due to an unexpected cwd.

clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The logic that finds an existing PR comment to update matches any comment whose body starts with the marker string, regardless of author. On public PRs, another user could post a comment beginning with the same marker and this workflow would overwrite it. Restrict the match to comments created by this workflow (e.g., `c.user.login === 'github-actions[bot]'`) and/or include an unguessable hidden marker in the comment body to identify it uniquely before updating.
2. `scripts/validate-bibliography-references.py`: `_find_repo_root()` docstring says the function falls back to `here.parent` "with a warning", but the current implementation returns the fallback silently. Either emit a warning (e.g., to stderr) when the `.git` marker isn’t found, or update the docstring so it matches the actual behavior.

clarkemoyer:

1. `scripts/validate-bibliography-references.py`: The per-page status label prints `PASS` whenever there are no *unmatched* refs, even if there are *possible* (fuzzy) matches. That makes the console output misleading (a page with only fuzzy matches is not fully validated). Consider treating `possible` as a warning (or using a third status like `MAYBE`) so operators can distinguish clean passes from pages needing manual verification.

2. `scripts/validate-bibliography-references.py`: When `get_zotero_client()` exits (e.g. missing env vars / missing dependency), `main()` writes an error report with the fixed message "Could not connect to Zotero", which loses the actionable reason that operators need to fix CI. Consider including the underlying cause (e.g. the missing variable name(s) or the ImportError message / exit context) in the JSON `error` field so the workflow PR comment is diagnostic.


clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The Markdown escaping helper doesn’t escape `&`, which allows entity-based HTML injection (e.g. `&lt;...&gt;` becomes a real tag after rendering). Because this runs in a `pull_request_target` context and embeds untrusted strings (from page content / exception messages) into a PR comment, `escMd` should escape `&` (typically first) to prevent entity smuggling.
2. `.github/workflows/validate-bibliography-zotero.yml`: The overlay step deletes *all* `src/app/bibliography-*/page.tsx` files from the base checkout before checking out the PR-head versions. If the PR branch is behind `main` (or the PR-head simply doesn’t contain newly-added bibliography pages), those pages will be removed from the workspace even though they would still exist after merge, causing the validator to under-count pages/refs and potentially miss unmatched citations. Consider overlaying PR-head files onto the base checkout without a blanket delete, and only removing files explicitly deleted in the PR (e.g. via `git diff --name-status base..pr-head`).

clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: The overlay logic doesn’t handle RENAMED bibliography pages correctly: `git diff --diff-filter=AMR --name-only` will check out the new path, but the old (pre-rename) `src/app/bibliography-*/page.tsx` from the base branch will remain because it isn’t listed under `--diff-filter=D`. This causes the validator to run against both old and new pages (double-counting refs / reporting stale unmatched refs) even though the rename would remove the old page on merge. Handle renames explicitly (e.g., use `git diff --name-status --diff-filter=R` to remove the old path, or use `--diff-filter=DR` with name-status parsing) before overlaying the renamed file.
2. `.github/workflows/validate-bibliography-zotero.yml`: Use `python -m pip install ...` (as done in other workflows here) instead of invoking `pip` directly. This avoids accidentally installing into a different Python environment than the one configured by `actions/setup-python`, which can cause intermittent CI failures depending on runner PATH state.


clarkemoyer:

1. `.github/workflows/validate-bibliography-zotero.yml`: `git merge-base HEAD pr-head` in the overlay step relies on having enough git history to compute a merge base. With the default `actions/checkout` shallow clone, `merge-base` can fail or return incorrect results, which would break the diff-based overlay logic (and could lead to validating the wrong file set). Consider setting `fetch-depth: 0` (or otherwise ensuring sufficient history) on the initial checkout so merge-base/diff computations are reliable.


clarkemoyer:

1. `scripts/validate-bibliography-references.py`: The comment says the explicit `None` check makes an empty-string top-level key defer to `data.key`, but the implementation (`_raw_key if _raw_key is not None else ...`) will *not* defer when `_raw_key == ''`. Either adjust the comment to match current behavior, or change the logic to treat empty-string like missing so the fallback is actually used.

PR #1696, opened 2026-04-17
crp-tools: bring offline CRP document tools under version control

clarkemoyer:
Refs #1693.

Adds `scripts/crp-document-tools/` with the full set of offline tools used to author, validate, and cross-check the TABS CRP document and its appendixes. These tools were previously versioned only on the author's local workspace with EST-timestamped filenames; this PR establishes proper version control in the public repo so every revision has a commit trail.

## Why a new subtree rather than `scripts/analysis/`

`scripts/analysis/` runs in CI and produces `src/data/*.json` that the website consumes. Those scripts are the **live pipeline**, publicly reproducible.

`scripts/crp-document-tools/` operates on the CRP `.docx`, the appendix markdown files, and the pipeline JSON. It is **not** run in CI today because it requires a local copy of the private CRP workspace on disk. Keeping the two subtrees separate makes the runtime contract explicit: analysis pipeline = public + CI, document tools = local-only author-facing toolchain.

## Folder layout

```
scripts/crp-document-tools/
  README.md
  validators/       # Static assertion checks against the CRP docx
    validate_crp_stats_v5.py             # CURRENT (277 checks, 42 categories)
    validate_appendixes.py
    archive/                             # v1, v2, merged, v3, v4
  builders/         # Document construction
    merge_appendixes.py                  # CURRENT
    compute_crp_stats_v2.py
    archive/                             # earlier merge variant
  convergence/      # CRP <-> pipeline <-> website reconciliation
    crp_full_convergence_check_v2_3.py   # CURRENT
    crp_convergence_check.py             # baseline simpler version
    archive/                             # four prior v2 iterations
```

## All validator versions preserved

All six `validate_crp_stats` revisions are included. `v5` is current at the top of `validators/`; `v1` through `v4` plus the legacy `merged` variant are under `validators/archive/`. Going forward, new validator work happens on branches against `validators/validate_crp_stats_v5.py` (or a successor) and archive copies accumulate naturally in git history - no more copying files with EST-timestamped filenames.

## Privacy verification

Scanned every added file. No PII, no tokens, no secrets, no api_keys. Only survey role-label mappings (e.g., 'CEO (e.g., Agency Director, ...)' -> 'CEO') and workspace path globs (`/sessions/*/mnt/! Clarke Moyer Smeal CRP - TABS`, `/tmp/tabs-site`). These are runtime-discoverable paths, not sensitive data.

All 20 added files were copied byte-for-byte from their workspace origins with no content edits (only filename normalization for the convergence archive entries that had spaces and parentheses in their workspace names).

## Dash hygiene

Zero em or en dashes in any of the three new README files (scanned with a Python check). Only ASCII hyphen-minus and U+2212 Unicode minus where appropriate.

## What this PR does not do

- Does not wire any of the new tools into CI (would require either a docx fixture or a stripped-down CRP test document)
- Does not delete the workspace copies; the author continues to edit locally and push revisions through normal git flow going forward
- Does not move the CRP body .docx source into the repo (would require committee approval on data access policy)

## Future direction

1. Add a CI workflow that runs `validate_crp_stats` on PRs touching `src/data/crp-*.json`, using a fixture docx
2. Add a pre-commit hook running the minimal convergence check
3. Consider moving CRP body .docx source into the repo behind a data access policy

## Test plan

- [x] No em or en dashes in any new file
- [x] No PII in any script (grep scan for PROLIFIC_PID, Auth_LLM, Auth_Bots, email, token, secret, api_key)
- [x] All 20 files copied byte-for-byte from workspace (no content edits)
- [x] README files use CommonMark with blank lines before lists
- [ ] Manual review of folder layout and README clarity
- [ ] Verify at least one script still runs against the local workspace after being pulled from the repo clone


clarkemoyer:

1. `scripts/crp-document-tools/builders/merge_appendixes.py`: `CRP_ROOT`/`CRP_PATH` are derived from the script’s location (repo checkout), so the script looks for the body `.docx` inside `scripts/crp-document-tools/builders/` and for appendixes in `scripts/crp-document-tools/02 CRP Appendixes/`. Those inputs are expected to live in the external CRP workspace, so running this from the repo root will not find any of the required files. Make the workspace/docx/appendix paths configurable (CLI args and/or glob-based workspace discovery like `validators/validate_crp_stats_v5.py`) and resolve inputs from the workspace’s `01 CRP Body/` + `02 CRP Appendixes/` folders.

2. `scripts/crp-document-tools/builders/merge_appendixes.py`: `_find_appendix()` picks the “latest” appendix by lexicographically sorting filenames. With timestamps like `4-9-2026` vs `4-16-2026`, lexicographic order can select an older file as “latest” (because `'9' > '1'`). Parse the timestamp in the filename into a datetime (or sort by file mtime) to reliably pick the newest appendix.

3. `scripts/crp-document-tools/validators/validate_appendixes.py`: `CRP_ROOT` is hard-coded to a specific `/sessions/...` workspace path. Other “current” tools (e.g., `validate_crp_stats_v5.py`) auto-discover the workspace via glob or accept CLI args, which makes them portable across sessions/machines. Consider aligning this script by adding workspace discovery and/or a `--workspace` argument, then deriving `APPENDIX_DIR` from that.
4. `scripts/crp-document-tools/validators/validate_appendixes.py`: `discover_appendix_files()` relies on lexicographic filename sorting and overwrites `APPENDIX_FILES[...]` as it iterates, which can select the wrong “latest” appendix when timestamps aren’t zero-padded (e.g., `4-9-2026` vs `4-16-2026`). Prefer parsing the timestamp into a datetime (or selecting by mtime) and explicitly choosing the newest match per appendix.
5. `scripts/crp-document-tools/validators/validate_appendixes.py`: Appendix markdown is opened without an explicit encoding. Since the appendix files may contain non-ASCII characters (smart quotes, Unicode minus, etc.), this can break on systems with a non-UTF-8 default locale. Open the markdown with `encoding='utf-8'` (or `utf-8-sig` if needed) for consistent behavior.

6. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: This script writes outputs under `args.output_dir/Pipeline Snapshots`, `Claims Databases`, and `Convergence Reports` but never creates those directories. A first-time run will raise `FileNotFoundError` on the `open(..., 'w')` calls. Create the required subdirectories with `os.makedirs(..., exist_ok=True)` before writing the files.

7. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `ts_file` is computed but never used. Either remove it or use it consistently for filenames (if the intent is to avoid spaces/colons in output paths).
8. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: If `clean_rows` is empty, `durations` will be empty and this line will crash (indexing, division by zero, `min`/`max`). Guard the duration summary with a non-empty check and print a clear message when there are no clean rows / no durations available.

9. `scripts/crp-document-tools/builders/merge_appendixes.py`: `pytz` is imported/used for timestamps here, but it’s not listed in the builders’ documented dependencies. Either document `pytz` as a required dependency, or switch to the stdlib `zoneinfo` (Python 3.9+) to avoid the extra runtime requirement.

10. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: `CSV_PATH` is hard-coded to a specific `/sessions/.../mnt/uploads/...` location, which makes the script non-portable and contradicts the READMEs’ “auto-discover via glob” expectation. Prefer a `--csv` CLI arg (with a sensible default/discovery routine), and consider defaulting to the public de-identified dataset path when available.

11. `scripts/crp-document-tools/builders/README.md`: The dependency list says `pandas` is used for CSV reading in `compute_crp_stats_v2.py`, but that script currently uses the stdlib `csv` module and doesn’t import `pandas`. Update the dependency list to match reality (or switch the script to actually use pandas if that’s the intent).

12. `scripts/crp-document-tools/README.md`: The privacy note says PII-containing/enriched CSVs are “NEVER referenced by these scripts”, but at least `validators/validate_crp_stats_v5.py` explicitly tries to load `*Enriched_CRP200*.csv` before falling back to the public dataset. Please reconcile the documentation with the actual behavior (e.g., clarify that enriched/local-only CSVs may be used but must never be committed, and that the public dataset is the safe default).


clarkemoyer:

1. `scripts/crp-document-tools/builders/merge_appendixes.py`: Several imported symbols appear unused (`Inches`, `Cm`, `Emu`, and `etree`). Removing them reduces implied dependencies and avoids lint/IDE warnings.

2. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: This archived validator contains a hard-coded Qualtrics `ResponseId` (`R_1QK12IJpHjC3wd6`). Even in an archive script, committing respondent identifiers is risky from a privacy standpoint. Please remove/redact the identifier (or replace with a non-identifying comment explaining the historical exception) before merging.

3. `scripts/crp-document-tools/convergence/crp_full_convergence_check_v2_3.py`: Like the minimal convergence checker, this uses a fixed UTC-5 offset but labels timestamps as “EST”; that will be wrong during daylight saving time, and `%-m`/`%-d` formatting is also non-portable. Consider switching to `zoneinfo.ZoneInfo('America/New_York')` and/or omitting the zone label to avoid misleading timestamps.

4. `scripts/crp-document-tools/validators/validate_appendixes.py`: There are unused imports here (`json` and `Path`). Dropping unused imports keeps these scripts easier to audit and avoids implying extra dependencies.

5. `scripts/crp-document-tools/validators/validate_appendixes.py`: The “Validation check count (CRP)” consistency check is hard-coded to expect 84 checks. This looks out of date with the v5 validator (277 checks) referenced elsewhere in this subtree; consider updating the expected value/pattern (or allowing both) so the report stays meaningful.
6. `scripts/crp-document-tools/builders/merge_appendixes.py`: Unused imports (`copy` here) add noise and can confuse readers about what functionality is required. Please remove if not needed.
7. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: This script hard-codes a fixed UTC-5 offset but labels timestamps as “EST”. That will be incorrect during daylight saving time. Consider using `zoneinfo.ZoneInfo('America/New_York')` (as in merge_appendixes.py) and formatting the zone name accurately (or omitting the zone label).

8. `scripts/crp-document-tools/validators/validate_appendixes.py`: The ImportError guidance suggests installing pypdf with `--break-system-packages`, which is distro-specific and can damage system Python installs. Prefer recommending a virtualenv/pipx (or just `pip install pypdf` inside a venv) and keep the message platform-neutral.

9. `scripts/crp-document-tools/validators/validate_appendixes.py`: `extract_page_claims()` is currently dead code and its `label_pattern` parameter is unused. Either remove it (and related unused helpers) or wire it into the report so it’s clear what it’s for.

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: Avoid bare `except:` here; it will swallow non-PDF parsing errors (including KeyboardInterrupt/SystemExit) and can hide real issues in link counting. Catch `Exception` explicitly and consider collecting/reporting the exception type/count so failures are diagnosable.
2. `scripts/crp-document-tools/builders/merge_appendixes.py`: `strftime('%-m-%-d-...')` is not portable (e.g., it fails on Windows). Since these are local tools, it’s still worth using a platform-safe timestamp format (e.g., using `now.month`/`now.day` like the convergence scripts do, or `%m/%d` with stripping) so the builder can run in more environments.

3. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: This archived validator hard-codes absolute, session-specific `CSV_PATH`/`DOCX_PATH` and immediately opens them at import time. As committed, it’s effectively non-runnable outside the original session path and will fail with an unhelpful stack trace. Consider refactoring to accept `--csv/--docx` (or workspace discovery like later versions) and wrapping execution in a `main()` guarded by `if __name__ == '__main__':` so it’s safer to reference/import.
4. `scripts/crp-document-tools/validators/README.md`: The archive table says `validate_crp_stats_v1.py` is “Initial release, 23 checks”, but the script currently contains many more check calls than 23. Please update this note to reflect the actual scope (or describe it approximately) so readers don’t underestimate what v1 covers.

5. `scripts/crp-document-tools/validators/validate_appendixes.py`: `expected_refs` is defined as a set of tuples, so iteration order is non-deterministic. That makes the report output order unstable between runs and also treats the “could cross-reference …” entries the same as required references (status becomes `MISSING`). Consider switching this to an ordered list and explicitly encoding whether each reference is required vs optional (so optional ones are reported as INFO/OPTIONAL rather than missing).
6. `scripts/crp-document-tools/validators/validate_appendixes.py`: Cross-reference detection uses case-sensitive substring checks (`p in text`). This will miss variants like `APPENDIX C`, `Appendix c.1`, or other casing/formatting differences. Use a case-insensitive search (e.g., lowercase both sides or regex with `re.IGNORECASE`) so the validator doesn’t produce false MISSING results.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: Avoid bare `except:` here; it will swallow non-PDF parsing errors (including KeyboardInterrupt/SystemExit) and can hide real issues in link counting. Catch `Exception` explicitly and consider collecting/reporting the exception type/count so failures are diagnosable.
2. `scripts/crp-document-tools/builders/merge_appendixes.py`: `strftime('%-m-%-d-...')` is not portable (e.g., it fails on Windows). Since these are local tools, it’s still worth using a platform-safe timestamp format (e.g., using `now.month`/`now.day` like the convergence scripts do, or `%m/%d` with stripping) so the builder can run in more environments.

3. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: This archived validator hard-codes absolute, session-specific `CSV_PATH`/`DOCX_PATH` and immediately opens them at import time. As committed, it’s effectively non-runnable outside the original session path and will fail with an unhelpful stack trace. Consider refactoring to accept `--csv/--docx` (or workspace discovery like later versions) and wrapping execution in a `main()` guarded by `if __name__ == '__main__':` so it’s safer to reference/import.
4. `scripts/crp-document-tools/validators/README.md`: The archive table says `validate_crp_stats_v1.py` is “Initial release, 23 checks”, but the script currently contains many more check calls than 23. Please update this note to reflect the actual scope (or describe it approximately) so readers don’t underestimate what v1 covers.

5. `scripts/crp-document-tools/validators/validate_appendixes.py`: `expected_refs` is defined as a set of tuples, so iteration order is non-deterministic. That makes the report output order unstable between runs and also treats the “could cross-reference …” entries the same as required references (status becomes `MISSING`). Consider switching this to an ordered list and explicitly encoding whether each reference is required vs optional (so optional ones are reported as INFO/OPTIONAL rather than missing).
6. `scripts/crp-document-tools/validators/validate_appendixes.py`: Cross-reference detection uses case-sensitive substring checks (`p in text`). This will miss variants like `APPENDIX C`, `Appendix c.1`, or other casing/formatting differences. Use a case-insensitive search (e.g., lowercase both sides or regex with `re.IGNORECASE`) so the validator doesn’t produce false MISSING results.


clarkemoyer:
**Copilot Review Cycle (Round 3/7):** Fix timeout after retry. Dispatching next round to re-check. 6 comment(s) may still need attention.

clarkemoyer:

1. `scripts/crp-document-tools/builders/archive/merge_appendixes_v2.py`: Even as an archive script, this hard-codes a session-specific absolute workspace path and timestamped filenames. That makes the script non-runnable for anyone else and contradicts the general “workspace discovery / no session-specific paths” approach used elsewhere in this subtree. Consider either (a) converting these to CLI args + discovery like the current builder, or (b) adding an explicit note at the top that this archived variant is intentionally non-portable and kept for reference only.
2. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: `check()` is nested inside `main()`, but it declares `global checks, passed, failed` while `checks/passed/failed` are defined as locals in `main()`. This will raise at runtime (the `global` variables don't exist at module scope). Use `nonlocal` here, or move the counters to module scope / avoid the nested function so the counters are updated correctly.

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: The header says "Python 3.8+" but this script imports `zoneinfo.ZoneInfo`, which is only available in the stdlib on Python 3.9+. Please update the stated requirement (or provide a backport/fallback) so users don't hit an import error.

4. `scripts/crp-document-tools/builders/merge_appendixes.py`: `add_appendix_title_page()` calls `docx.enum.text.WD_BREAK.PAGE`, but the `docx` package itself is only imported later via `import docx.enum.text`, and the function returns a `(title_p, None)` tuple that the caller ignores. Consider importing what you need at the top (e.g., `import docx` or `from docx.enum.text import WD_BREAK`) and returning only the paragraph to avoid relying on import side effects and unused return values.
5. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: The v5 module docstring mixes current and historical details in a way that's misleading for users (e.g., the "Usage" line still references `validate_crp_stats_v4.py`, and the stated total assertion count (~400) conflicts with `validators/README.md` describing 277 checks). Please update the docstring so the run command and check count match the current v5 behavior, and keep the v4 header clearly separated if you want to preserve it verbatim.
6. `scripts/crp-document-tools/README.md`: The runtime requirements list is incomplete: `validate_crp_stats_v5.py` imports `scipy` (and `validate_appendixes.py` requires `pypdf`), but neither is listed here. Consider adding the full set of dependencies (and/or noting which sub-tools require which packages) to prevent setup surprises.


clarkemoyer:

1. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `crp_convergence_check.py` never assigns the status `MISMATCH` to any claim, but the report/exit code logic keys off `status_counts['MISMATCH']`. As a result, the script can exit 0 and print “All claims converge” even when claims remain `UNVERIFIED`, which is misleading for a convergence gate. Consider either (a) explicitly setting `MISMATCH` for remaining unclassified claims after `classify_mismatches`, or (b) treating `UNVERIFIED` as a non-zero exit condition / reporting it in the mismatch section.

2. `scripts/crp-document-tools/convergence/crp_full_convergence_check_v2_3.py`: The registry filename suffix stripping regex is hard-coded to `... EST)` but `get_timestamp()` uses `%Z` in the America/New_York timezone, which will produce `EDT` during daylight saving time. When registry files are timestamped with EDT, this regex won’t strip the suffix and the derived `key` won’t match expected registry names. Update the regex to accept both `EST` and `EDT` (or any `[A-Z]{3}`), or force the timestamp format to a consistent fixed zone label.

3. `scripts/crp-document-tools/validators/validate_appendixes.py`: `extract_capture_page_claims_from_appendix()` is currently unused in this script. If it’s not part of the intended validation flow, consider removing it to reduce maintenance surface area; if it is intended, wire it into the report so it actually contributes findings.

4. `scripts/crp-document-tools/validators/validate_appendixes.py`: The filename timestamp parser only matches `(... EST)`; if appendix filenames are generated during daylight saving time using `%Z` (e.g., `EDT`), `_parse_filename_date()` will fail and return `(0,0,0,0)`, breaking “latest file” selection. Consider accepting `EST|EDT` (or any timezone abbreviation) in the regex, or standardizing the file naming to a fixed label.

clarkemoyer:

1. `scripts/crp-document-tools/builders/README.md`: This README repeatedly claims filenames are “EST-timestamped”, but the builder uses `ZoneInfo('America/New_York')`, which yields `EDT` during DST. As written, users running in April/September will produce `(... EDT)` filenames, contradicting the docs.

Consider updating the documentation to say “ET (EST/EDT)” (or similar), or change the script to always emit a fixed `EST` label if that’s truly required by the project’s naming rule.
2. `scripts/crp-document-tools/convergence/README.md`: The “Running” section shows both convergence scripts being invoked with no CLI arguments, but `crp_convergence_check.py` (and likely `crp_full_convergence_check_v2_3.py`) requires `--crp-xml`, `--repo`, and `--output-dir`. As-is, copying these commands will fail immediately.

Either update the README examples to include the required flags, or add the same workspace/repo auto-discovery behavior the other tool subtrees use so the no-arg invocation is actually valid.

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: The docstring claims outputs are named with underscores (e.g., `pipeline_stats_<timestamp>.json`), but the implementation writes files with spaces and parentheses (e.g., `pipeline_stats (<timestamp>).json`). This mismatch makes it harder to locate outputs and can confuse users reading the script header.

Update the docstring to match the actual filenames (or rename the outputs to match what the docstring promises).

4. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `generate_report()` builds a “MISMATCHES REQUIRING ATTENTION” section by filtering `status == 'MISMATCH'`, but no code path ever sets a claim’s status to `MISMATCH` (claims remain `UNVERIFIED` or are reclassified as `TIER_MATCH` / `ITEM_LEVEL` / etc.). As a result, the report will always say “No Mismatches Found” even when there are unmatched claims.

Either introduce a real `MISMATCH` classification for claims that don’t match any pipeline value, or change the report to surface `UNVERIFIED` (and/or other non-converged statuses) in the “requires attention” section.
5. `scripts/crp-document-tools/builders/merge_appendixes.py`: `_parse_filename_date()` only matches timestamps ending in `EST`, but `merge_appendixes.py` later generates output filenames using `%Z` from `America/New_York`, which will be `EDT` during DST. That means date parsing (and therefore “latest appendix” selection) can silently fail for `(... EDT)` filenames and fall back to `(0,0,0,0)` ordering.

Update the regex to accept both `EST` and `EDT` (or any 3-letter TZ), or normalize the timestamp label used in filenames so parsing and generation are consistent.


clarkemoyer:
**Copilot Review Cycle (Round 6/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:
**Copilot Review Cycle (Round 7/7):** Max rounds reached. 5 comment(s) still need human attention.

clarkemoyer:

1. `scripts/crp-document-tools/builders/merge_appendixes.py`: `add_formatted_runs()` uses the `re` module (`re.split`, `re.match`), but this file never imports `re`, so `merge_appendixes.py` will crash with a `NameError` the first time it processes markdown content. Add the missing import (or avoid `re` usage) so the script is runnable.
2. `scripts/crp-document-tools/convergence/README.md`: This README says the full convergence check “auto-discovers workspace via glob” and that paths are auto-discovered from `/sessions/*/...`, but `crp_full_convergence_check_v2_3.py` (and `crp_convergence_check.py`) require explicit `--crp-xml/--repo/--output-dir` (and `--registry-dir`) and contain no workspace discovery logic. Please update the wording here to match how the scripts actually locate inputs.
3. `scripts/crp-document-tools/validators/validate_appendixes.py`: This script bypasses the detailed error handling in `paths.find_workspace()` by (a) taking `args.workspace` as a raw string and (b) using a local `find_workspace()` wrapper that swallows `CrpWorkspaceNotFound` and returns `None`. As a result, an invalid `--workspace` path produces a generic “Could not locate appendix directory” error instead of the specific message from `paths.py`, and the discovery precedence described in the module docstring isn’t consistently enforced. Consider using the shared helpers directly (e.g., `_find_workspace_shared(explicit=args.workspace)` and/or `paths.find_appendix_dir(...)`) and letting `CrpWorkspaceNotFound` surface its actionable message.
4. `.github/workflows/crp-tools-smoke.yml`: These bash blocks use `set -uo pipefail` but not `-e`. In particular, `out=$(python ...)` will not fail the step if the script exits non-zero (as long as the subsequent marker greps succeed), which can mask real failures. Use `set -euo pipefail` (or explicitly check the Python exit code) so CI reliably fails on non-zero exits.

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: `discover_appendix_files()` only recognizes appendix markdown filenames starting with "Appendix A/B/C/D" (and "Appendix Project Plan"), but the bundled fixture + subtree README describe the appendix files as `A_...md`, `B_...md`, etc. With the current logic the validator will report all appendixes as NOT FOUND and then run most checks against empty text, which can produce a misleading "clean" report. Update discovery to also accept the `A_`/`B_`/`C_`/`D_` naming scheme (or align the fixture/docs to the `Appendix X` prefix) and fail fast if required appendixes are missing.
2. `scripts/crp-document-tools/builders/merge_appendixes.py`: Appendix discovery here hard-requires filenames starting with `Appendix A` / `Appendix B` / etc. The subtree README + the bundled fixture appendixes are named `A_TABS_...md`, `B_TABS_...md`, etc, so `merge_appendixes.py` will exit with "No file found" when run against the fixture or a workspace following the documented layout. Consider accepting both naming conventions (e.g., try `Appendix {letter}` then `{letter}_`) or update the fixture/docs to match the enforced prefix.

3. `scripts/crp-document-tools/fixtures/build_example_workspace.py`: The generated stub appendix H1s are missing a space after the hyphen (e.g. "# Appendix A -Survey Instrument"). This looks like a typo in the fixture content and will keep regenerating on every fixture rebuild. Adjust the stub strings to "# Appendix A - Survey Instrument" (and similarly for B/C/D) so the example workspace reads cleanly.

4. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `extract_pipeline_stats()` selects `val['samples'][0]`, but the validation JSON schema also includes `primary_sample` (key) and the sample ordering is not guaranteed. Prefer selecting the sample matching `primary_sample` (fall back to samples[0] only if it’s missing) so convergence checks stay correct if additional samples/tiers are added or reordered.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: The numeric regex patterns in `check_number_consistency()` can produce false positives because they don't enforce digit boundaries (e.g., `39\s+workflows` will also match the substring "39 workflows" inside "139 workflows"). Tighten these patterns with digit boundaries like `(?<!\d)` / `(?!\d)` (or `\b` where safe) so counts are not overstated.

2. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: `load_survey()` only validates `REQUIRED_COLS`, but later code assumes many additional columns exist (all `BARRIER_COLS`/`READINESS_COLS`/`MATURITY_COLS` plus the three IRI columns). If the CSV schema is missing any of these, the script will raise a `KeyError`/`IndexError` with a stack trace instead of a clear error. Consider validating presence of all referenced survey columns up front and raising a descriptive `ValueError` that lists any missing item/IRI columns.

3. `scripts/crp-document-tools/fixtures/example_workspace/02 CRP Appendixes/A_TABS_Survey_Instrument_(4-16-2026 0109 EDT).md`: This fixture stub heading is missing a space after the hyphen (`"-Survey"`). Also, it does not match the heading text written by `fixtures/build_example_workspace.py` (which includes `"- Survey"`), so re-running the fixture builder will rewrite this file and leave the repo dirty. Align the committed stub content with the generator output (or adjust the generator) so the fixture is idempotent.

4. `scripts/crp-document-tools/fixtures/example_workspace/02 CRP Appendixes/B_TABS_Research_Platform_(4-16-2026 0054 EDT).md`: This fixture stub heading is missing a space after the hyphen (`"-Research"`). Also, it does not match the heading text written by `fixtures/build_example_workspace.py` (which includes `"- Research"`), so re-running the fixture builder will rewrite this file and leave the repo dirty. Align the committed stub content with the generator output (or adjust the generator) so the fixture is idempotent.

5. `scripts/crp-document-tools/fixtures/example_workspace/02 CRP Appendixes/C_Data_Analysis_and_Validation_(4-16-2026 0054 EDT).md`: This fixture stub heading is missing a space after the hyphen (`"-Data"`). Also, it does not match the heading text written by `fixtures/build_example_workspace.py` (which includes `"- Data"`), so re-running the fixture builder will rewrite this file and leave the repo dirty. Align the committed stub content with the generator output (or adjust the generator) so the fixture is idempotent.

6. `scripts/crp-document-tools/fixtures/example_workspace/02 CRP Appendixes/D_Institutional_Governance_(4-16-2026 0054 EDT).md`: This fixture stub heading is missing a space after the hyphen (`"-Institutional"`). Also, it does not match the heading text written by `fixtures/build_example_workspace.py` (which includes `"- Institutional"`), so re-running the fixture builder will rewrite this file and leave the repo dirty. Align the committed stub content with the generator output (or adjust the generator) so the fixture is idempotent.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: find_latest_docx() picks the “latest” file via lexicographic sort (sorted(...)[-1]). With filenames that embed non-zero-padded timestamps like "(4-9-2026 ...)" vs "(4-16-2026 ...)", lexicographic ordering can select an older document as the newest. Use the shared helpers in paths.py (e.g., find_body_docx()/parse_filename_date with mtime fallback) so date-aware selection is correct and consistent across the toolchain.
2. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: find_csv() also selects a CSV via sorted(matches)[-1], which has the same lexicographic timestamp pitfall as the docx selection (single-digit month/day can sort after double-digit). Prefer paths.find_survey_csv() (mtime-based) or a date-aware parse_filename_date+mtime strategy so the validator consistently targets the newest intended CSV.
3. `.github/workflows/crp-tools-smoke.yml`: PR description says the new crp-document-tools are *not* wired into CI, but this workflow does run a fixture-driven smoke test on PRs and pushes. Either update the PR description (or README wording) to reflect that a minimal CI smoke run is now in place, or adjust the workflow scope if it’s not intended to be part of CI.

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: `validate_appendixes.py` re-implements filename timestamp parsing in `_parse_filename_date`, but the subtree already provides `paths.parse_filename_date()` (more permissive TZ token and shared semantics). Duplicating this logic risks the appendix/capture “latest file” selection drifting from the rest of the toolchain; consider importing and using `parse_filename_date` from `paths.py` instead of maintaining a local variant here.
2. `scripts/crp-document-tools/builders/merge_appendixes.py`: `_find_appendix()` selects the “latest” candidate solely via `parse_filename_date()`. If multiple candidates lack an embedded timestamp (so `parse_filename_date` returns `(0,0,0,0)`), `max()` will pick an arbitrary file (depends on glob order). To make selection deterministic in that case, include a secondary key (e.g., file mtime) similar to `find_latest_docx()`.

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `deduplicate_claims()` dedupes by `(crp_value, char_offset//500)` only. This can incorrectly drop distinct claims that share the same numeric value within the same ~500-char window (e.g., different metrics that coincidentally equal 0.85 in one paragraph), reducing report coverage and potentially hiding real mismatches. Include additional discriminators (at least category/subcategory and/or the matched text span) in the dedupe key so only true duplicates are removed.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: The number consistency checks hard-code a 24-character hex "Prolific Study ID" value. Even if this is not a participant PID, it matches the repo’s PID shape and is easy to confuse with prohibited identifiers; it may also get flagged by automated scans. Consider masking it (e.g., only last 6 chars), moving it to a local-only config, or using a looser regex that validates format + surrounding prose instead of embedding the exact ID in version control.

2. `scripts/crp-document-tools/validators/validate_appendixes.py`: The Appendix A inline page-claim scan will treat any match whose nearby context contains the word "Capture" as a Survey Instrument page-count claim and compare it against `survey_instrument` pages. Because "Capture" is a generic term, this can produce false mismatches for other captures mentioned in Appendix A. Tighten the condition to only trigger when the match is clearly tied to the Survey Instrument capture (e.g., require "Survey Instrument" in the matched window, or key off the capture’s configured short_name).

3. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: `pearson()` can raise ZeroDivisionError when there are fewer than 2 valid (a,b) pairs (it divides by `n` and `n-1` without guarding). This is reachable if one of the per-respondent mean lists ends up empty after filtering. Add an `if n < 2: return None` (or similar) early, and consider handling zero-variance cases as well to avoid crashes on edge-case CSVs.

clarkemoyer:

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: In discover_appendix_files(), selection uses only parse_filename_date() as the max() key. When appendix filenames don’t contain a timestamp (parse_filename_date returns (0,0,0,0)), the “latest” file becomes dependent on os.listdir() order and can be nondeterministic. Consider falling back to file mtime (or a secondary sort key) when the parsed timestamp is missing, consistent with paths.parse_filename_date()’s contract and other callers in this subtree.
2. `scripts/crp-document-tools/builders/merge_appendixes.py`: When --workspace is provided, the script doesn’t validate that it exists/is a directory; it just treats it as truthy and then fails later with the generic "No CRP body DOCX found. Provide --workspace or --docx." This is misleading for a typo’d workspace path and makes debugging harder. Consider using the shared paths.find_workspace(explicit=...) behavior (or an explicit os.path.isdir check) so an invalid --workspace produces a clear, immediate error.

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: In _resolve_capture_file(), the glob-fallback selects the “newest” candidate using parse_filename_date() only. If any candidate lacks a parseable timestamp (parse returns (0,0,0,0)), max() will fall back to glob order rather than actual recency, which can silently pick the wrong capture. Consider using a secondary sort key (e.g., os.path.getmtime) when parse_filename_date() returns the zero tuple, similar to the appendix discovery logic.

2. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: The module docstring says the original session-specific paths were removed, but find_workspace() still relies on /sessions/* glob patterns. Either update the docstring to reflect that this archive version still uses legacy session globs, or refactor it to delegate to scripts/crp-document-tools/paths.py for portable discovery.

3. `scripts/crp-document-tools/validators/README.md`: The “Running” section claims validate_crp_stats_v5.py auto-discovers the workspace “via glob”, but the current script delegates discovery to scripts/crp-document-tools/paths.py (which uses --workspace/env/user locations/fixture/legacy globs in a defined precedence). Updating this README to describe the actual paths.py precedence will avoid confusing users when glob-only discovery doesn’t match their environment.


clarkemoyer:

1. `scripts/crp-document-tools/paths.py`: `find_body_docx` uses `workspace = workspace or find_workspace()`, so a typo’d/missing `--workspace` passed through callers won’t be validated via `find_workspace(explicit=...)` (and won’t get the explicit-override error messaging described in this module). Consider routing the optional `workspace` argument through `find_workspace(explicit=workspace)` (and applying the same pattern to `find_appendix_dir` / `find_survey_csv`) so explicit overrides are consistently validated.
2. `scripts/crp-document-tools/validators/validate_appendixes.py`: The Prolific Study ID regex appears off by one: `[0-9a-f]{18}ead2da5` matches 25 hex chars total (18 + 7), but the comment says this is a 24-char hex ID. As written, this check will never match and will always report the study ID as missing. Adjust the quantifier so the total length is 24 (or otherwise encode the intended length explicitly).


clarkemoyer:

1. `scripts/crp-document-tools/paths.py`: In `find_workspace()`, the legacy fallback iterates `_legacy_glob_locations()` in ascending `sorted(glob)` order and returns the first match. If multiple `/sessions/*` mounts exist, this can pick an arbitrary/older workspace and make local runs non-deterministic. Consider choosing the most recent candidate (e.g., `max(..., key=os.path.getmtime)` across all legacy matches) or at least returning the last sorted entry to preserve historical behavior.
2. `.github/workflows/crp-tools-smoke.yml`: The determinism check only hashes the synthetic CSV, but `build_example_workspace.py` also rewrites the appendix stub markdown files and the body README. If any future change introduces nondeterminism in those outputs, this workflow would still pass. Consider hashing all files under `fixtures/example_workspace/` (or at least the generated stub files too) to fully assert fixture rebuild determinism.


clarkemoyer:

clarkemoyer:

1. `.github/workflows/crp-tools-smoke.yml`: The fixture determinism hash step uses `find ... | sort | xargs sha256sum`, but the fixture tree contains paths with spaces (e.g., `05 TABS Survey Support/...`). `xargs` will split on whitespace and break hashing (or hash the wrong paths). Use a NUL-delimited pipeline (`find -print0 | sort -z | xargs -0 sha256sum`) or another approach that preserves spaces reliably.

2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `extract_pipeline_stats()` is looking for reliability data under lowercase keys (`'barriers'/'readiness'/'maturity'`) and fields like `alpha`, but `src/data/crp-validation.json` stores these under capitalized construct keys (`"Barriers"`, `"Readiness"`, `"Maturity"`) and fields like `cronbach_alpha`/`mcdonalds_omega`/etc. As written, this will silently skip most pipeline stats and cause many claims to be reported as UNVERIFIED. Update the extractor to support the actual schema (and ideally fall back to both key styles).

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `extract_pipeline_stats()` populates Fornell–Larcker stats using `entry.get('shared_variance')`, but the current `crp-validation.json` schema's `fornell_larcker` entries expose fields like `sqrt_ave1`, `sqrt_ave2`, `abs_r`, and `pass` (no `shared_variance`). This will leave `fl_*` values as None and reduce convergence coverage. Adjust the extraction to align with the JSON schema (or compute shared variance from available fields if that's what the report expects).


clarkemoyer:

1. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: `compute_crp_stats_v2.py` relies on `paths.find_survey_csv()` with its default glob (`*Enriched_CRP200*.csv`). That means auto-discovery will fail for workspaces that only contain the public frozen dataset (e.g., `TABS_V2_CRP_2026_public_dataset.csv`) or the `*V2_ONLY*.csv` variant, even though those are valid inputs for this script. Consider mirroring `validate_crp_stats_v5.find_csv()` here: try a small ordered list of patterns and then fall back to the public dataset filename, so `--csv` isn’t required in common setups.

2. `scripts/crp-document-tools/validators/validate_appendixes.py`: `count_words()` tries to strip markdown table separators with `re.sub(r'\|[-:]+\|', '', text)`, but it won’t match common separator rows like `| --- | --- |` (spaces) and only partially matches multi-column separators. This will inflate word counts for table-heavy appendixes. Consider dropping entire table-separator lines (e.g., line-based filter) or using a regex that matches full separator rows including whitespace and multiple columns.
3. `scripts/crp-document-tools/validators/validate_appendixes.py`: `check_cross_references()` uses simple substring checks (e.g., searching for `c.1`) when validating sub-references. This can produce false positives (for example, a reference to `C.10` contains `c.1` as a substring). Using a regex with boundaries (e.g., `\bC\.1\b` / `\bAppendix\s+C\.1\b`, case-insensitive) would avoid incorrectly marking references as present.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: `validate_crp_stats_v5.py` prints FAIL/WARN results but still exits with status 0 as long as docx/csv discovery succeeds. That makes the validator unusable in automation (and contradicts the validators README which states non-zero exit when any check FAILs). Consider exiting non-zero when `validator.counts['FAIL'] > 0` (and optionally keep WARN non-fatal as documented).

2. `scripts/crp-document-tools/README.md`: The Quickstart table shows an example `python validate_crp_stats_v5.py --workspace ...`, but `validate_crp_stats_v5.py` does not currently accept a `--workspace` flag (only `--docx/--csv/--repo`). Either add `--workspace` support (and pass it through to `paths.find_workspace`) or adjust this doc/example so it matches the actual CLI.

3. `scripts/crp-document-tools/validators/README.md`: The discovery precedence comment mentions `--workspace <path>` as an explicit CLI flag for `validate_crp_stats_v5.py`, but that script currently has no `--workspace` argument. Also, this README states exit codes reflect FAIL status, but the script currently never sets a non-zero exit code when checks FAIL. Please align the README with the actual CLI/exit behavior, or update the script to match this documented contract.
4. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: `find_repo()` only checks a few hard-coded `/tmp/tabs-site*` paths. When running the documented command from this repo (e.g. `python scripts/crp-document-tools/validators/validate_crp_stats_v5.py`), `--repo` is typically omitted and the canonical pipeline JSON is available in the current checkout, but this function will return `None` and the validator won’t actually load `src/data/crp-*.json`. Consider adding a repo-root auto-detection (e.g. based on `__file__`/subtree location or `os.getcwd()` when `src/data/crp-validation.json` exists), or require/validate `--repo` explicitly.

clarkemoyer:

1. `scripts/crp-document-tools/paths.py`: `REPO_ROOT` is declared but never used in this module. Consider removing it, or using it to build repo-relative paths (or prefixing with `_` to make the intentional-unused status clear).

2. `scripts/crp-document-tools/paths.py`: `find_body_docx()` claims to fall back to file mtime when the filename lacks an embedded timestamp, but the current sort key makes any non-timestamped file sort behind timestamped files (because the parsed tuple starts with 0). Either adjust the selection logic (e.g., only consider timestamped files if any exist, otherwise use mtime; or convert mtime to a comparable date tuple) or clarify the docstring so callers understand the precedence.

3. `scripts/crp-document-tools/builders/merge_appendixes.py`: `find_latest_docx()` says it falls back to file mtime when a filename has no embedded timestamp, but returning `(0, 0, 0, mtime)` means non-timestamped files can never win if any timestamped file exists. Either make the fallback truly compete (e.g., normalize mtime into the same tuple shape) or clarify in the docstring that timestamped filenames always take precedence over mtime.
4. `scripts/crp-document-tools/convergence/README.md`: The README says the convergence tools produce "two artifacts", but `crp_convergence_check.py` (and the full checker) also writes a pipeline snapshot JSON per run. Update the list/count so it matches the actual outputs users will see in the output directory.

5. `scripts/crp-document-tools/convergence/README.md`: This line says prior versions are moved to `Old/`, but the current convergence scripts only write timestamped files; they don't appear to relocate older outputs. Either implement the move behavior or reword this as a manual/project convention rather than script behavior.

6. `scripts/crp-document-tools/builders/README.md`: This claims the builders move prior versions to `Old/`, but `merge_appendixes.py`/`compute_crp_stats_v2.py` currently just write new timestamped outputs without relocating older ones. Either implement the move-to-Old behavior or reword this as a manual/project convention.

7. `scripts/crp-document-tools/convergence/archive/crp_full_convergence_check_v2_1656.py`: `strftime("%-m-%-d-...")` is not portable (it breaks on Windows because `%-m`/`%-d` aren’t supported). Consider switching to a portable format (or constructing the string via `dt.month`/`dt.day` as v2.3 does) so archived scripts remain runnable cross-platform, or explicitly document the POSIX-only constraint.

8. `scripts/crp-document-tools/convergence/archive/crp_full_convergence_check_v2_1659.py`: `strftime("%-m-%-d-...")` is not portable (it breaks on Windows because `%-m`/`%-d` aren’t supported). Consider switching to a portable format (or constructing the string via `dt.month`/`dt.day` as v2.3 does) so archived scripts remain runnable cross-platform, or explicitly document the POSIX-only constraint.

9. `scripts/crp-document-tools/convergence/archive/crp_full_convergence_check_v2_1.py`: `strftime("%-m-%-d-...")` is not portable (it breaks on Windows because `%-m`/`%-d` aren’t supported). Consider switching to a portable format (or constructing the string via `dt.month`/`dt.day` as v2.3 does) so archived scripts remain runnable cross-platform, or explicitly document the POSIX-only constraint.

10. `scripts/crp-document-tools/convergence/archive/crp_full_convergence_check_v2_2.py`: `strftime("%-m-%-d-...")` is not portable (it breaks on Windows because `%-m`/`%-d` aren’t supported). Consider switching to a portable format (or constructing the string via `dt.month`/`dt.day` as v2.3 does) so archived scripts remain runnable cross-platform, or explicitly document the POSIX-only constraint.

11. `scripts/crp-document-tools/validators/archive/validate_crp_stats_v1.py`: The docstring says this workspace discovery uses the same pattern as v5, but this archived v1 script uses hard-coded `/sessions/*` globs (v5 uses `paths.py` with broader precedence rules). Updating the docstring would prevent confusion for anyone trying to run the archive script.


clarkemoyer:

1. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `extract_pipeline_stats()` returns early when `crp-validation.json` has no `samples`, which prevents loading `crp-sensitivity-analysis.json` even if it exists. This makes the convergence pass silently drop sensitivity-tier stats in scenarios like older/alternate validation schemas. Suggestion: avoid returning from inside the validation block; instead skip validation extraction and continue so sensitivity stats still load (and only return at the end).

2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: The module header says it “Extracts all quantitative claims” and “Cross-references every claim”, but the current regex/number parsing only covers a narrow subset (e.g., sample sizes only match digits with no thousands separators, and `parse_num()` won’t parse values like `1,234` or percentages). Either soften the wording to reflect the heuristic nature of the extractor, or extend the extraction/parsing to handle common CRP formats (commas, percents, etc.).
3. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: The discovery-precedence summary here is out of sync with `paths.py` (which checks both `~/Documents/CRP-workspace` and `~/CRP-workspace`, includes the repo fixture, and also supports legacy locations like `/tmp/tabs-crp-workspace`). Updating this docstring to match the actual precedence in `scripts/crp-document-tools/paths.py` will prevent confusion when users try to run the script without `--csv`/`--workspace`.

4. `scripts/crp-document-tools/validators/README.md`: This precedence list references `paths.py`, but it omits the `/tmp/tabs-crp-workspace` legacy fallback that `paths.py` will also consider. Suggestion: include that path (or rephrase this section to point readers to `paths.py` as the canonical source of truth) so the README stays accurate as discovery rules evolve.


clarkemoyer:

1. `scripts/crp-document-tools/builders/merge_appendixes.py`: merge_appendixes.py wraps paths.find_workspace() and swallows CrpWorkspaceNotFound, returning None. If CRP_WORKSPACE is set but points to a missing directory (or legacy discovery fails), this turns a clear discovery error into a later, misleading "No CRP body DOCX found" failure. Consider only skipping workspace discovery when --docx and --appendix-dir are provided, and otherwise propagate the discovery error so users see the actionable message from paths.py.

2. `scripts/crp-document-tools/convergence/README.md`: This README instructs users to run `unpack.py` to extract word/document.xml, but there is no unpack.py included in this repo. Either add the helper script under this subtree or update the docs to describe an existing, reproducible extraction method (e.g., unzipping the .docx and reading word/document.xml).

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: main() returns `mismatches + unverified` as the process exit code. Exit codes are typically treated as boolean success/failure (and may be truncated modulo 256 by shells), so returning a potentially large count can be misleading. Consider returning 0 when both counts are 0, otherwise return 1 (while still printing the counts).


clarkemoyer:

clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_crp_stats_v5.py`: find_workspace() swallows CrpWorkspaceNotFound and returns None. In main(), this causes the script to fall through to generic "Could not find CRP body .docx/CSV" errors instead of emitting the actionable workspace-discovery error that points to the Quickstart. Recommend letting CrpWorkspaceNotFound propagate (or printing its message and exiting) so missing-workspace failures are diagnosed correctly.

2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: The header comment references an `unpack.py` prerequisite, but there is no such script in this repo. This is likely stale/incorrect documentation; suggest either removing the reference or updating it to the documented unzip/Python snippet (or to the actual helper script name, if one exists).


clarkemoyer:

1. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `strip_xml_tags()`/`parse_num()` don’t normalize the Unicode minus sign (U+2212). Since the CRP doc uses U+2212 for negative numbers, negative claims (e.g., correlations) will fail `float()` parsing and silently drop out of the extracted-claims set. Consider normalizing `\u2212` to `-` (and optionally stripping thousands separators / trailing `%`) before numeric parsing so negative values are reliably extracted.
2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: The numeric regexes in `extract_claims()` are likely to miss common CRP formats that omit a leading zero (e.g., `r = .34`, `M = .86`) and any values written with Unicode minus (e.g., `r = −.21`). Updating the mean/correlation patterns (or normalizing the text first) to accept optional leading digits and both `-` and `\u2212` will materially improve coverage and reduce false `UNVERIFIED` findings.

clarkemoyer:

1. `scripts/crp-document-tools/paths.py`: `find_workspace()` returns `FIXTURE_DIR` before checking legacy `/sessions/*` globs. Because the fixture directory exists in every repo checkout, this makes the legacy auto-discovery branch effectively unreachable in default runs (breaking the stated goal of keeping legacy automation working unchanged). Consider moving the fixture fallback after legacy globs, or gating fixture selection behind an explicit flag/env var (e.g. `CRP_USE_FIXTURE=1` / `CI=true`).
2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `classify_mismatches()` changes `UNVERIFIED` claims to `TIER_MATCH`/`ITEM_LEVEL`/`SUBGROUP` purely based on context keywords, without actually matching the value to any pipeline statistic. Because the process exit code only fails on `UNVERIFIED`/`MISMATCH`, this can incorrectly produce a zero exit status (false convergence) for claims that don’t match the pipeline. Consider only reclassifying after confirming a numeric match to an appropriate pipeline key/tier, and otherwise keep the claim as `UNVERIFIED` (optionally adding a separate `hint` field for heuristic context).

3. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `MISMATCH` is described in the report and included in the “requires attention” filter, but this script never assigns `status='MISMATCH'` anywhere. As a result, the report semantics are misleading and the `mismatches` counter in `main()` will always be 0. Either implement a real `MISMATCH` classification for confirmed contradictions, or remove `MISMATCH` from the status model/report and rely on `UNVERIFIED` only.
4. `scripts/crp-document-tools/fixtures/build_example_workspace.py`: `build_example_workspace.py` enforces LF-only output for the CSV, but the appendix stub files and `01 CRP Body/README.txt` are written with default newline translation. On Windows, this can produce CRLF outputs and break the workflow’s determinism check (which runs on Linux and regenerates LF). Consider opening these text files with `newline='\n'` (or writing bytes) to make the fixture rebuild byte-identical across platforms.

clarkemoyer:
**Copilot Review Cycle (Round 3/7):** Fixes pushed but CI timed out (likely action_required gate). Dispatching next round automatically.

clarkemoyer:

1. `scripts/crp-document-tools/validators/README.md`: The workspace discovery precedence list here doesn’t match the actual precedence in scripts/crp-document-tools/paths.py. paths.py checks legacy /sessions/* (and /tmp/tabs-crp-workspace) *before* falling back to the repo fixture, but this README lists the fixture before legacy. Please update this list to match paths.py so users on machines with legacy mounts don’t get surprised by which workspace is chosen.

2. `scripts/crp-document-tools/README.md`: This paragraph says discovery falls back to the bundled fixture and “finally” to legacy /sessions paths, but paths.py actually tries legacy globs (and /tmp/tabs-crp-workspace) before using the fixture as a last resort. Please align this description with paths.py’s documented/implemented order to avoid misleading users about which workspace will be selected.

3. `scripts/crp-document-tools/builders/compute_crp_stats_v2.py`: The discovery precedence in this docstring is out of sync with scripts/crp-document-tools/paths.py: it lists the repo fixture before legacy /sessions/* globs, but paths.py prefers legacy locations before falling back to the fixture. Please update the order shown here so it accurately reflects runtime behavior.

4. `scripts/crp-document-tools/fixtures/build_example_workspace.py`: This docstring states “All text files are opened with newline='\n'”, but the CSV is opened with newline="" (and relies on csv.writer(lineterminator='\n') for deterministic bytes). Please adjust the wording to reflect what the script actually does so the determinism mechanism is clear and accurate.


clarkemoyer:

1. `scripts/crp-document-tools/validators/validate_appendixes.py`: The docstring’s workspace discovery precedence is inaccurate relative to `scripts/crp-document-tools/paths.py`. `paths.py` tries legacy `/sessions/*` + `/tmp/tabs-crp-workspace` before falling back to the bundled fixture, and it also includes `~/CRP-workspace` in common user locations. Please update this description to match the authoritative precedence to avoid users setting up the wrong paths and getting unexpected fixture fallbacks.

2. `scripts/crp-document-tools/fixtures/example_workspace/01 CRP Body/README.txt`: The instructions for generating a synthetic body `.docx` don’t specify that the command must be run from this `01 CRP Body/` directory (or otherwise save the file into this folder). As written, it’s easy to generate the docx in an arbitrary working directory, after which `merge_appendixes.py` won’t auto-discover it under `01 CRP Body/`. Please clarify the required working directory or provide an explicit output path in the command.


clarkemoyer:
**Copilot Review Cycle (Round 5/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/crp-document-tools/validators/README.md`: The workspace discovery precedence list implies `/sessions/*` is always preferred over `/tmp/tabs-crp-workspace`, but `find_workspace()` in `paths.py` treats both as legacy candidates and selects the most-recently-modified directory across *all* legacy matches. Update this README to match the actual behavior (or adjust `paths.py` if strict ordering is intended).

2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows installs that don’t ship an IANA tz database (unless the `tzdata` package is installed). If these tools are intended to be cross-platform, consider catching that exception and either (a) emitting a clear install hint for `tzdata`, or (b) falling back to `datetime.now().astimezone()` / UTC for timestamps.

3. `scripts/crp-document-tools/builders/merge_appendixes.py`: `ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows installs that don’t ship an IANA tz database (unless the `tzdata` package is installed). If these builders are intended to be cross-platform, catch that exception here and either emit a clear remediation (install `tzdata`) or fall back to a safe timezone strategy (e.g., local tz via `astimezone()` or UTC) for filename timestamps.
4. `scripts/crp-document-tools/convergence/crp_full_convergence_check_v2_3.py`: `NY = ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows systems without an IANA tz database (unless the `tzdata` package is installed). If this script is intended to be portable, wrap the ZoneInfo initialization in a try/except and provide a clear install hint for `tzdata` or a fallback timezone (e.g., local tz / UTC).


clarkemoyer:

1. `scripts/crp-document-tools/validators/README.md`: The workspace discovery precedence list implies `/sessions/*` is always preferred over `/tmp/tabs-crp-workspace`, but `find_workspace()` in `paths.py` treats both as legacy candidates and selects the most-recently-modified directory across *all* legacy matches. Update this README to match the actual behavior (or adjust `paths.py` if strict ordering is intended).

2. `scripts/crp-document-tools/convergence/crp_convergence_check.py`: `ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows installs that don’t ship an IANA tz database (unless the `tzdata` package is installed). If these tools are intended to be cross-platform, consider catching that exception and either (a) emitting a clear install hint for `tzdata`, or (b) falling back to `datetime.now().astimezone()` / UTC for timestamps.

3. `scripts/crp-document-tools/builders/merge_appendixes.py`: `ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows installs that don’t ship an IANA tz database (unless the `tzdata` package is installed). If these builders are intended to be cross-platform, catch that exception here and either emit a clear remediation (install `tzdata`) or fall back to a safe timezone strategy (e.g., local tz via `astimezone()` or UTC) for filename timestamps.
4. `scripts/crp-document-tools/convergence/crp_full_convergence_check_v2_3.py`: `NY = ZoneInfo('America/New_York')` can raise `ZoneInfoNotFoundError` on Windows systems without an IANA tz database (unless the `tzdata` package is installed). If this script is intended to be portable, wrap the ZoneInfo initialization in a try/except and provide a clear install hint for `tzdata` or a fallback timezone (e.g., local tz / UTC).


clarkemoyer:
**Copilot Review Cycle (Round 6/7):** Fix timeout after retry. Dispatching next round to re-check. 4 comment(s) may still need attention.

clarkemoyer:

clarkemoyer:

1. `scripts/crp-document-tools/fixtures/build_example_workspace.py`: `build_example_workspace.py` regenerates `example_workspace/01 CRP Body/README.txt`, but the string written here does not match the committed fixture README content. As a result, the CI step "Verify the fixture rebuilds deterministically" will detect a byte-level change after regeneration and fail. Update `write_body_placeholder()` so it writes the exact committed README.txt (or update the committed README.txt to match the generator) to keep rebuilds idempotent.

PR #1701, opened 2026-04-17
fix(crp-2026/top-barriers): only show omission note when an item is actually omitted

clarkemoyer:
## Problem

On `/results/crp-2026/top-barriers` the interpretation paragraph below the Pick-vs-Mean table ended with:

> Barriers omitted from the Mean Rank column are those with no valid continuous-rating mean in the frozen dataset.

Readers look for an omission that never happens. In the frozen CRP N=200 dataset every one of the 18 substantive barriers has a valid continuous-rating mean (per-item N range 198 to 200), so zero barriers are ever missing from the Mean Rank column.

## Fix

- Keep the defensive `filter` that excludes items with null means (so a future partial-response case would still render correctly).
- Compute `hasOmittedMeanRank = pickSorted.some((r) => meanRankOf[r.item] === undefined)` at module scope.
- Only render the omission sentence when `hasOmittedMeanRank` is true.

With the current frozen data, readers now see only the Delta interpretation, without the dangling reference to an omission that does not exist.

## Verification

All 18 barriers have valid means in `src/data/crp-sensitivity-analysis.json` (verified via the pipeline-generated JSON). `npx tsc --noEmit` and `npx eslint` both pass clean.

### DIFF
--- a/src/app/results/crp-2026/top-barriers/page.tsx
+++ b/src/app/results/crp-2026/top-barriers/page.tsx
@@ -62,7 +62,7 @@ const pickRankOf: Record<string, number> = {}
 pickSorted.forEach((r, i) => {
   pickRankOf[r.item] = i + 1
 })
-const meanRankOf: Record<string, number> = {}
+const meanRankOf: Partial<Record<string, number>> = {}
 meanSorted.forEach((r, i) => {
   meanRankOf[r.item] = i + 1
 })
@@ -372,8 +372,8 @@ const TopBarriersPage = () => {
             participants must prioritize), and a positive Delta means the opposite (it falls under
             forced choice relative to continuous rating).
             {hasOmittedMeanRank
-              ? ' Barriers missing a Mean Rank value above are those with no valid continuous-rating mean in this dataset (all responses missing for that item).'
-              : ''}
+              ? ' Barriers missing a Mean Rank value above are those with no valid continuous-rating responses for that item in this dataset.'
+              : null}
           </p>
         </section>
 
--- a/src/app/results/crp-2026/top-barriers/page.tsx
+++ b/src/app/results/crp-2026/top-barriers/page.tsx
@@ -157,16 +157,18 @@ const positiveMovers = [...moversAll].filter((r) => r.delta > 0).sort((a, b) => 
 const topNegative = negativeMovers.slice(0, 2)
 const topPositive = positiveMovers.slice(0, 2)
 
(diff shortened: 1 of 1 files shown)


clarkemoyer:

1. `src/app/results/crp-2026/top-barriers/page.tsx`: `hasOmittedMeanRank` becomes `true` whenever `pickSorted` is non-empty but `meanRankOf` is empty (e.g., when the pipeline didn’t write `item_descriptives` / there are no numeric means at all). In that situation the conditional sentence implies “all responses missing for that item”, but the real issue is that the mean data block is unavailable. Consider gating this with mean data presence (e.g., `meanRankable.length > 0 && ...`) or otherwise ensuring the omission note can’t appear when mean ranks are entirely unavailable.
```suggestion
// every response is missing. Only show the explanatory sentence when mean-rank data exists and
// there is an actual omission for at least one picked item.
const hasOmittedMeanRank =
  meanRankable.length > 0 && pickSorted.some((r) => meanRankOf[r.item] === undefined)
```

clarkemoyer:

1. `src/app/results/crp-2026/top-barriers/page.tsx`: `meanRankOf` is typed as `Record<string, number>`, but this code relies on missing keys returning `undefined` (items without valid means are intentionally absent). Consider changing the type to reflect that (e.g., `Partial<Record<string, number>>` / `Record<string, number | undefined>`), so the `=== undefined` check and other `meanRankOf[r.item] ?? null` usages are type-accurate and safer to refactor.
```suggestion
const meanRankLookup: Partial<Record<string, number>> = meanRankOf
const hasOmittedMeanRank =
  meanRankable.length > 0 && pickSorted.some((r) => meanRankLookup[r.item] === undefined)
```
2. `src/app/results/crp-2026/top-barriers/page.tsx`: The conditional omission sentence claims this happens when "all responses" are missing for an item, but the pipeline can emit `mean=null` when all responses are blank, "Don't Know", or otherwise unmapped (i.e., not necessarily missing). Please adjust the wording to match the actual data condition: no *valid* continuous-rating responses for that item.
```suggestion
              ? ' Barriers missing a Mean Rank value above are those with no valid continuous-rating responses for that item in this dataset.'
```
3. `src/app/results/crp-2026/top-barriers/page.tsx`: In JSX, using `: ''` renders an empty text node; prefer returning `null`/`undefined` for the false branch so nothing is rendered when there’s no omission note.
```suggestion
              : null}
```

clarkemoyer:

Issue #1780, opened 2026-04-23
feat: inline glossary popovers for psychometric terms on results pages


clarkemoyer:
## User-facing problem

Results pages (`/results/*`, `/results/crp-2026/*`) are dense with psychometric terms — KMO, HTMT, Cronbach's α, CFI, RMSEA, composite reliability, invariance, etc. Even though `/results/glossary` exists, readers have to tab away and hunt for the term, which breaks flow and filters out non-academic audiences.

## What

A reusable `<Term>` component that wraps a term inline and shows its definition in a tooltip/popover on hover (desktop) or tap (mobile). Keyboard-focusable for a11y; tooltip closes on blur / outside tap.

```tsx
<Term term="KMO">KMO</Term>
// renders the word "KMO" with an accessible tooltip showing the
// definition sourced from the same glossary that powers /results/glossary
```

## Why

- Non-academic readers (execs, practitioners, the CMO Survey's broader audience) give up on pages full of undefined jargon.
- `/results/glossary` already exists — we have the content, just not the UX to surface it in context.
- Compounds across ~15 results pages with no per-page content rewrite.

## Acceptance criteria

- [ ] `<Term>` component exists in `src/components/glossary-term/` with proper a11y (focusable, `role="tooltip"`, ESC dismisses, works on touch)
- [ ] Glossary data is the single source of truth — popover text and `/results/glossary` both read from the same structure
- [ ] Terms wrapped on: `/results/reliability`, `/results/factor-analysis`, `/results/validation`, `/results/crp-2026/reliability`, `/results/crp-2026/factor-analysis`, `/results/crp-2026/validation` (at minimum)
- [ ] jest-axe passes on a page with the component used
- [ ] Unit tests: opens on focus, closes on blur/ESC, reads definition from glossary data
- [ ] Playwright test: tap / hover on term reveals popover

## Not in scope

- Adding new terms to the glossary (content work, separate)
- Rewriting existing copy (we only wrap existing jargon)
- Server-side rendering of tooltips (client-side is fine for a static export)

## Size

~1-2 hours of work. Single component + data wiring + mechanical wrapping of terms on ~5 pages.

PR #1782, opened 2026-04-23
feat(glossary): inline <Term> tooltip component for psychometric terms

clarkemoyer:
Partially addresses #1780 — the `<Term>` component is fully implemented and wired on both reliability pages; remaining pages are tracked as follow-ups below.

## Summary

Adds an inline `<Term>` component that shows a glossary definition in a popover on hover / focus / tap for terms like Cronbach's α, KMO, HTMT, CFI, RMSEA, AVE. Keeps readers in the flow of the results pages instead of sending them to `/results/glossary` and back.

## What changed

- `src/data/glossary-terms.ts` (new) — single source of truth for inline popover copy. 25 entries extracted from the existing `/results/glossary` page.
- `src/components/glossary-term/` (new) — the `<Term termId="...">` component.
  - Renders as a `<button>` with a dotted underline
  - Opens on hover, focus, or tap; closes on blur / Esc / click-outside
  - **`role="dialog"`** (not `role="tooltip"`) — the popover contains a focusable "See full entry" link, which ARIA 1.2 forbids inside `role="tooltip"`. This is intentional.
  - `aria-haspopup="dialog"`, `aria-expanded`, `aria-controls`, and `aria-describedby` wired in
  - Click **toggles** open/closed when activated by keyboard or touch; click is a no-op when the mouse is hovering (hover already controls open state, preventing a hover+click race that would immediately close the popover)
  - Falls back to plain text when `termId` isn't in the glossary (typo-safe)
- `__tests__/components/glossary-term/` (new) — 16 tests including two jest-axe a11y checks (open and closed), toggle-on-click, and hover-suppresses-toggle. All pass.
- Wired Cronbach's alpha on `/results/reliability` and `/results/crp-2026/reliability` to demonstrate the pattern.

## Test plan

- [x] 16/16 unit tests pass (`jest __tests__/components/glossary-term`)
- [x] jest-axe passes in both closed and open states
- [x] `npm run lint` — 0 errors
- [x] `npm run build` — both reliability pages prerender
- [ ] CI green
- [ ] Playwright: hover on Cronbach's alpha on `/results/reliability` reveals popover; Esc closes

## Follow-up scope (future PRs)

- Wrap additional terms on `/results/factor-analysis`, `/results/validation`, and `/results/crp-2026/*` counterparts (mechanical)
- Consider a `<TermList>` bibliography-style index for pages with many terms

### DIFF
--- a/__tests__/components/glossary-term/index.test.tsx
+++ b/__tests__/components/glossary-term/index.test.tsx
@@ -78,6 +78,23 @@ describe('Term', () => {
     expect(screen.getByText('nonexistent-term-xyz')).toBeInTheDocument()
   })
 
+  it('closes when a pointerdown event fires outside the component', () => {
+    render(
+      <div>
+        <Term termId="kmo">KMO</Term>
+        <button type="button" data-testid="outside">
+          Outside
+        </button>
+      </div>
+    )
+    // Open the tooltip
+    fireEvent.click(screen.getAllByRole('button')[0])
+    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    // Pointer down outside the component closes it
+    fireEvent.pointerDown(screen.getByTestId('outside'))
+    expect(screen.queryByRole('tooltip')).toBeNull()
+  })
+
   it('has no accessibility violations in the closed state', async () => {
     const { container } = render(<Term termId="kmo">KMO</Term>)
     const results = await axe(container)
--- a/__tests__/components/glossary-term/index.test.tsx
+++ b/__tests__/components/glossary-term/index.test.tsx
@@ -55,6 +55,32 @@ describe('Term', () => {
     expect(screen.queryByRole('dialog')).toBeNull()
   })
 
+  it('toggles closed on a second click (keyboard/touch — no mouse hover active)', () => {
+    render(<Term termId="rmsea">RMSEA</Term>)
+    const btn = screen.getByRole('button')
+
+    // First click opens (no mouseenter, so mouseInsideRef is false → toggle)
+    fireEvent.click(btn)
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+
+    // Second click closes (mouseInsideRef still false → toggle again)
+    fireEvent.click(btn)
+    expect(screen.queryByRole('dialog')).toBeNull()
+  })
+
+  it('does not toggle closed on click when mouse is inside (hover-open scenario)', () => {
+    render(<Term termId="rmsea">RMSEA</Term>)
+    const btn = screen.getByRole('button')
+
+    // Simulate hover opening the popover
+    fireEvent.mouseEnter(btn)
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+
+    // Click while mouse is still inside should NOT close the popover
+    fireEvent.click(btn)
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+  })
+
   it('sets aria-expanded in sync with open state and opens a dialog', () => {
     render(<Term termId="htmt">HTMT</Term>)
     const btn = screen.getByRole('button')
--- a/__tests__/components/glossary-term/index.test.tsx
+++ b/__tests__/components/glossary-term/index.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { render, screen, fireEvent } from '@testing-library/react'
+import { render, screen, fireEvent, act, waitFor } from '@testing-library/react'
 import { axe, toHaveNoViolations } from 'jest-axe'
 import Term from '@/components/glossary-term'
 
@@ -16,14 +16,14 @@ describe('Term', () => {
     expect(screen.getByRole('button')).toHaveTextContent('KMO')
   })
 
-  it('opens the tooltip on click and renders the short definition', () => {
+  it('opens the dialog on click and renders the short definition', () => {
     render(<Term termId="kmo">KMO</Term>)
     const btn = screen.getByRole('button')
-    expect(screen.queryByRole('tooltip')).toBeNull()
+    expect(screen.queryByRole('dialog')).toBeNull()
     fireEvent.click(btn)
-    const tip = screen.getByRole('tooltip')
-    expect(tip).toBeInTheDocument()
-    expect(tip).toHaveTextContent(/sampling adequacy for factor analysis/i)
+    const dialog = screen.getByRole('dialog')
+    expect(dialog).toBeInTheDocument()
+    expect(dialog).toHaveTextContent(/sampling adequacy for factor analysis/i)
   })
 
   it('includes a link to the full glossary entry when open', () => {
@@ -42,29 +42,50 @@ describe('Term', () => {
     )
     const termBtn = screen.getAllByRole('button')[0]
     fireEvent.focus(termBtn)
-    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
     fireEvent.blur(termBtn)
-    expect(screen.queryByRole('tooltip')).toBeNull()
+    expect(screen.queryByRole('dialog')).toBeNull()
   })
 
   it('closes on Escape key', () => {
     render(<Term termId="rmsea">RMSEA</Term>)
     fireEvent.click(screen.getByRole('button'))
-    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
     fireEvent.keyDown(document, { key: 'Escape' })
-    expect(screen.queryByRole('tooltip')).toBeNull()
+    expect(screen.queryByRole('dialog')).toBeNull()
   })
 
-  it('sets aria-expanded and aria-describedby in sync with open state', () => {
+  it('sets aria-expanded in sync with open state and opens a dialog', () => {
     render(<Term termId="htmt">HTMT</Term>)
     const btn = screen.getByRole('button')
     expect(btn).toHaveAttribute('aria-expanded', 'false')
-    expect(btn).not.toHaveAttribute('aria-describedby')
+    expect(screen.queryByRole('dialog')).toBeNull()
 
     fireEvent.click(btn)
     expect(btn).toHaveAttribute('aria-expanded', 'true')
-    const tip = screen.getByRole('tooltip')
-    expect(btn).toHaveAttribute('aria-describedby', tip.getAttribute('id')!)
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+  })
+
+  it('restores focus to the trigger when Escape is pressed from within the dialog', async () => {
+    render(<Term termId="kmo">KMO</Term>)
+    const btn = screen.getByRole('button')
+
+    // Open the dialog
+    fireEvent.click(btn)
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+
+    // Move focus to the "See full entry" link inside the dialog
+    const link = screen.getByRole('link', { name: /see full entry/i })
+    act(() => link.focus())
+    expect(link).toHaveFocus()
+
+    // Press Escape — dialog should close and focus should return to trigger
+    fireEvent.keyDown(document, { key: 'Escape' })
+    expect(screen.queryByRole('dialog')).toBeNull()
+
+    await waitFor(() => {
+      expect(btn).toHaveFocus()
+    })
   })
 
   it('falls back to plain text when the termId is not in the glossary', () => {
@@ -87,12 +108,12 @@ describe('Term', () => {
         </button>
       </div>
     )
-    // Open the tooltip
+    // Open the dialog
     fireEvent.click(screen.getAllByRole('button')[0])
-    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
     // Pointer down outside the component closes it
     fireEvent.pointerDown(screen.getByTestId('outside'))
-    expect(screen.queryByRole('tooltip')).toBeNull()
+    expect(screen.queryByRole('dialog')).toBeNull()
   })
 
   it('has no accessibility violations in the closed state', async () => {
--- a/__tests__/components/glossary-term/index.test.tsx
+++ b/__tests__/components/glossary-term/index.test.tsx
@@ -1,0 +1,93 @@
+import React from 'react'
+import { render, screen, fireEvent } from '@testing-library/react'
+import { axe, toHaveNoViolations } from 'jest-axe'
+import Term from '@/components/glossary-term'
+
+expect.extend(toHaveNoViolations)
+
+describe('Term', () => {
+  it('renders the entry term by default', () => {
+    render(<Term termId="cronbach-alpha" />)
+    expect(screen.getByRole('button')).toHaveTextContent(/Cronbach/i)
+  })
+
+  it('accepts override children as the visible text', () => {
+    render(<Term termId="kmo">KMO</Term>)
+    expect(screen.getByRole('button')).toHaveTextContent('KMO')
+  })
+
+  it('opens the tooltip on click and renders the short definition', () => {
+    render(<Term termId="kmo">KMO</Term>)
+    const btn = screen.getByRole('button')
+    expect(screen.queryByRole('tooltip')).toBeNull()
+    fireEvent.click(btn)
+    const tip = screen.getByRole('tooltip')
+    expect(tip).toBeInTheDocument()
+    expect(tip).toHaveTextContent(/sampling adequacy for factor analysis/i)
+  })
+
+  it('includes a link to the full glossary entry when open', () => {
+    render(<Term termId="kmo">KMO</Term>)
+    fireEvent.click(screen.getByRole('button'))
+    const link = screen.getByRole('link', { name: /see full entry/i })
+    expect(link).toHaveAttribute('href', '/results/glossary#kmo')
+  })
+
+  it('opens on focus and closes on blur when focus leaves entirely', () => {
+    render(
+      <div>
+        <Term termId="htmt">HTMT</Term>
+        <button type="button">After</button>
+      </div>
+    )
+    const termBtn = screen.getAllByRole('button')[0]
+    fireEvent.focus(termBtn)
+    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    fireEvent.blur(termBtn)
+    expect(screen.queryByRole('tooltip')).toBeNull()
+  })
+
+  it('closes on Escape key', () => {
+    render(<Term termId="rmsea">RMSEA</Term>)
+    fireEvent.click(screen.getByRole('button'))
+    expect(screen.getByRole('tooltip')).toBeInTheDocument()
+    fireEvent.keyDown(document, { key: 'Escape' })
+    expect(screen.queryByRole('tooltip')).toBeNull()
+  })
+
+  it('sets aria-expanded and aria-describedby in sync with open state', () => {
+    render(<Term termId="htmt">HTMT</Term>)
+    const btn = screen.getByRole('button')
+    expect(btn).toHaveAttribute('aria-expanded', 'false')
+    expect(btn).not.toHaveAttribute('aria-describedby')
+
+    fireEvent.click(btn)
+    expect(btn).toHaveAttribute('aria-expanded', 'true')
+    const tip = screen.getByRole('tooltip')
+    expect(btn).toHaveAttribute('aria-describedby', tip.getAttribute('id')!)
+  })
+
+  it('falls back to plain text when the termId is not in the glossary', () => {
+    render(<Term termId="nonexistent-term-xyz">fallback text</Term>)
+    expect(screen.queryByRole('button')).toBeNull()
+    expect(screen.getByText('fallback text')).toBeInTheDocument()
+  })
+
+  it('renders the termId when a missing term has no children', () => {
+    render(<Term termId="nonexistent-term-xyz" />)
+    expect(screen.getByText('nonexistent-term-xyz')).toBeInTheDocument()
+  })
+
+  it('has no accessibility violations in the closed state', async () => {
+    const { container } = render(<Term termId="kmo">KMO</Term>)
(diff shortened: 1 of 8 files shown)


clarkemoyer:

1. `src/data/glossary-terms.ts`: The `as const` assertion is effectively neutralized here because the array is already type-annotated as `readonly GlossaryTermEntry[]` (so literal `id` values are widened to `string`). If the intent is to preserve literal ids (e.g., to derive a `GlossaryTermId` union), consider switching to `satisfies readonly GlossaryTermEntry[]` (and removing the explicit array annotation) or remove the `as const` to avoid implying stronger typing than is actually present.
2. `src/data/glossary-terms.ts`: The `as const` assertion is effectively neutralized here because the array is already type-annotated as `readonly GlossaryTermEntry[]` (so literal `id` values are widened to `string`). If the intent is to preserve literal ids (e.g., to derive a `GlossaryTermId` union), consider switching to `satisfies readonly GlossaryTermEntry[]` (and removing the explicit array annotation) or remove the `as const` to avoid implying stronger typing than is actually present.
3. `src/components/glossary-term/index.tsx`: Using `React.ReactNode` requires the `React` namespace type to be available in this module. To make this robust across TS/ESM configurations (and consistent with modern React typing), prefer `import type { ReactNode } from 'react'` and then `children?: ReactNode`.
4. `src/components/glossary-term/index.tsx`: Using `React.ReactNode` requires the `React` namespace type to be available in this module. To make this robust across TS/ESM configurations (and consistent with modern React typing), prefer `import type { ReactNode } from 'react'` and then `children?: ReactNode`.
5. `src/components/glossary-term/index.tsx`: The PR description says the tooltip closes on “click-outside” for tap as well, but the outside handler is wired to `mousedown` only, which can be unreliable/inconsistent for touch input. Use `pointerdown` (preferred) or add a `touchstart` listener alongside `mousedown` to ensure click-outside-to-close works on mobile/touch devices.
```suggestion
    const onClickOutside = (e: PointerEvent) => {
      if (!containerRef.current?.contains(e.target as Node)) {
        setOpen(false)
      }
    }
    document.addEventListener('keydown', onKey)
    document.addEventListener('pointerdown', onClickOutside)
    return () => {
      document.removeEventListener('keydown', onKey)
      document.removeEventListener('pointerdown', onClickOutside)
```
6. `src/components/glossary-term/index.tsx`: `buttonRef` is currently not referenced anywhere besides being passed to the `ref` prop. If there’s no planned focus/measurement use, remove it to reduce dead code; otherwise, add the intended usage (e.g., focusing the trigger after dismiss) to justify keeping it.
7. `src/components/glossary-term/index.tsx`: The click-outside-to-close behavior is newly introduced but not directly covered by tests. Add a unit test that opens the tooltip, clicks outside the component, and asserts the tooltip closes (and, if you switch to `pointerdown`, test that path as well).

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: Escape handling assumes the trigger still has focus, but focus can move into the tooltip (the “See full entry” link is focusable and the onBlur handler keeps the tooltip open in that case). If Escape is pressed while focus is on that link, the tooltip unmounts and focus will be lost (activeElement is removed). Consider restoring focus to the trigger when closing from within the tooltip (and preventing an immediate re-open on focus).
```suggestion
  const closeTooltip = ({ restoreFocus = false }: { restoreFocus?: boolean } = {}) => {
    const container = containerRef.current
    const trigger = container?.querySelector('button')
    const activeElement = document.activeElement
    const shouldRestoreFocus =
      restoreFocus &&
      trigger instanceof HTMLButtonElement &&
      !!activeElement &&
      container?.contains(activeElement) &&
      activeElement !== trigger

    setOpen(false)

    if (!shouldRestoreFocus || !(trigger instanceof HTMLButtonElement)) {
      return
    }

    const swallowProgrammaticFocus = (event: FocusEvent) => {
      event.stopPropagation()
      trigger.removeEventListener('focusin', swallowProgrammaticFocus, true)
    }

    trigger.addEventListener('focusin', swallowProgrammaticFocus, true)
    requestAnimationFrame(() => {
      trigger.focus()
    })
  }

  useEffect(() => {
    if (!open) return
    const onKey = (e: KeyboardEvent) => {
      if (e.key === 'Escape') {
        closeTooltip({ restoreFocus: true })
      }
    }
    const onClickOutside = (e: PointerEvent) => {
      if (!containerRef.current?.contains(e.target as Node)) {
        closeTooltip()
```
2. `src/components/glossary-term/index.tsx`: The popover includes a focusable Next.js <Link> inside an element with role="tooltip". Per ARIA tooltip patterns, tooltips should be non-interactive; interactive content should use a dialog/popover pattern instead. Consider either removing the link from the tooltip or switching to a popover/dialog semantics (e.g., role="dialog" with appropriate aria-labelledby/aria-describedby and focus management).
3. `src/data/glossary-terms.ts`: This PR claims glossary-terms.ts is the single source of truth for glossary copy, but /src/app/results/glossary/page.tsx currently defines its own ENTRIES array (including overlapping ids like cronbach-alpha, kmo, etc.). That means glossary content is duplicated and can drift. Consider refactoring so /results/glossary derives its entries from this data file (or vice versa) to keep one canonical source.
4. `src/components/glossary-term/index.tsx`: There is already a tooltip implementation in src/components/ui/tooltip.tsx; this introduces a second tooltip/popover pattern with separate event handling and positioning logic. To avoid duplicated tooltip behavior and future divergence, consider building <Term> on top of the shared Tooltip component or extracting common behavior into a shared utility/component.
5. `src/app/results/reliability/page.tsx`: This change adds new interactive hover/focus/tap behavior on a production results page, but there is no Playwright coverage exercising the tooltip behavior (open on hover/tap and close on Esc/click-outside). Since the repo already has Playwright E2E coverage, add a small spec that visits /results/reliability and verifies the Cronbach’s alpha tooltip opens and can be dismissed.

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: Escape handling assumes the trigger still has focus, but focus can move into the tooltip (the “See full entry” link is focusable and the onBlur handler keeps the tooltip open in that case). If Escape is pressed while focus is on that link, the tooltip unmounts and focus will be lost (activeElement is removed). Consider restoring focus to the trigger when closing from within the tooltip (and preventing an immediate re-open on focus).
```suggestion
  const closeTooltip = ({ restoreFocus = false }: { restoreFocus?: boolean } = {}) => {
    const container = containerRef.current
    const trigger = container?.querySelector('button')
    const activeElement = document.activeElement
    const shouldRestoreFocus =
      restoreFocus &&
      trigger instanceof HTMLButtonElement &&
      !!activeElement &&
      container?.contains(activeElement) &&
      activeElement !== trigger

    setOpen(false)

    if (!shouldRestoreFocus || !(trigger instanceof HTMLButtonElement)) {
      return
    }

    const swallowProgrammaticFocus = (event: FocusEvent) => {
      event.stopPropagation()
      trigger.removeEventListener('focusin', swallowProgrammaticFocus, true)
    }

    trigger.addEventListener('focusin', swallowProgrammaticFocus, true)
    requestAnimationFrame(() => {
      trigger.focus()
    })
  }

  useEffect(() => {
    if (!open) return
    const onKey = (e: KeyboardEvent) => {
      if (e.key === 'Escape') {
        closeTooltip({ restoreFocus: true })
      }
    }
    const onClickOutside = (e: PointerEvent) => {
      if (!containerRef.current?.contains(e.target as Node)) {
        closeTooltip()
```
2. `src/components/glossary-term/index.tsx`: The popover includes a focusable Next.js <Link> inside an element with role="tooltip". Per ARIA tooltip patterns, tooltips should be non-interactive; interactive content should use a dialog/popover pattern instead. Consider either removing the link from the tooltip or switching to a popover/dialog semantics (e.g., role="dialog" with appropriate aria-labelledby/aria-describedby and focus management).
3. `src/data/glossary-terms.ts`: This PR claims glossary-terms.ts is the single source of truth for glossary copy, but /src/app/results/glossary/page.tsx currently defines its own ENTRIES array (including overlapping ids like cronbach-alpha, kmo, etc.). That means glossary content is duplicated and can drift. Consider refactoring so /results/glossary derives its entries from this data file (or vice versa) to keep one canonical source.
4. `src/components/glossary-term/index.tsx`: There is already a tooltip implementation in src/components/ui/tooltip.tsx; this introduces a second tooltip/popover pattern with separate event handling and positioning logic. To avoid duplicated tooltip behavior and future divergence, consider building <Term> on top of the shared Tooltip component or extracting common behavior into a shared utility/component.
5. `src/app/results/reliability/page.tsx`: This change adds new interactive hover/focus/tap behavior on a production results page, but there is no Playwright coverage exercising the tooltip behavior (open on hover/tap and close on Esc/click-outside). Since the repo already has Playwright E2E coverage, add a small spec that visits /results/reliability and verifies the Cronbach’s alpha tooltip opens and can be dismissed.

clarkemoyer:
**Copilot Review Cycle (Round 2/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: The trigger button isn’t programmatically associated with the popover content. The dialog element has an `id={popoverId}`, but the trigger doesn’t reference it via `aria-controls`/`aria-describedby`, which can make the expanded content harder for assistive tech to discover. Consider wiring `aria-controls={popoverId}` (and keep `aria-expanded` tied to state).
2. `src/components/glossary-term/index.tsx`: PR description says the popover uses `role="tooltip"` and `aria-describedby`, but the implementation explicitly uses `role="dialog"` (and doesn’t set `aria-describedby`). Please align the PR description (and/or component behavior) so the intended a11y contract is accurate.
3. `src/app/results/glossary/page.tsx`: `const base = _termBase.get(extra.id)!` will throw at runtime if a PAGE_EXTRAS id is missing from `glossaryTerms` (or if ids drift during future edits), breaking the entire glossary page render. Prefer an explicit guard with a clear error (or filter/validate at module init) so missing ids fail fast with an actionable message rather than a non-null assertion crash.
```suggestion
  const base = _termBase.get(extra.id)

  if (!base) {
    throw new Error(
      `Glossary configuration error: PAGE_EXTRAS entry "${extra.id}" does not exist in glossaryTerms.`,
    )
  }
```
4. `src/components/glossary-term/index.tsx`: The popover is closed immediately on `onMouseLeave` for the trigger button. Because the popover is positioned with a `mb-2` gap (`bottom-full ... mb-2`), moving the pointer from the trigger to the popover can pass through empty space and cause the popover to unmount before the pointer reaches it, making the “See full entry” link hard/impossible to click. Consider handling hover at the container level and/or adding a small close delay (similar to `src/components/ui/tooltip.tsx`), and avoid closing on mouseleave while focus is still within the component.

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: The trigger button isn’t programmatically associated with the popover content. The dialog element has an `id={popoverId}`, but the trigger doesn’t reference it via `aria-controls`/`aria-describedby`, which can make the expanded content harder for assistive tech to discover. Consider wiring `aria-controls={popoverId}` (and keep `aria-expanded` tied to state).
2. `src/components/glossary-term/index.tsx`: PR description says the popover uses `role="tooltip"` and `aria-describedby`, but the implementation explicitly uses `role="dialog"` (and doesn’t set `aria-describedby`). Please align the PR description (and/or component behavior) so the intended a11y contract is accurate.
3. `src/app/results/glossary/page.tsx`: `const base = _termBase.get(extra.id)!` will throw at runtime if a PAGE_EXTRAS id is missing from `glossaryTerms` (or if ids drift during future edits), breaking the entire glossary page render. Prefer an explicit guard with a clear error (or filter/validate at module init) so missing ids fail fast with an actionable message rather than a non-null assertion crash.
```suggestion
  const base = _termBase.get(extra.id)

  if (!base) {
    throw new Error(
      `Glossary configuration error: PAGE_EXTRAS entry "${extra.id}" does not exist in glossaryTerms.`,
    )
  }
```
4. `src/components/glossary-term/index.tsx`: The popover is closed immediately on `onMouseLeave` for the trigger button. Because the popover is positioned with a `mb-2` gap (`bottom-full ... mb-2`), moving the pointer from the trigger to the popover can pass through empty space and cause the popover to unmount before the pointer reaches it, making the “See full entry” link hard/impossible to click. Consider handling hover at the container level and/or adding a small close delay (similar to `src/components/ui/tooltip.tsx`), and avoid closing on mouseleave while focus is still within the component.

clarkemoyer:
**Copilot Review Cycle (Round 3/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `src/app/results/glossary/page.tsx`: `_termBase.get(extra.id)!` will throw at runtime if `PAGE_EXTRAS` ever contains an id that’s missing from `glossaryTerms` (e.g., a future edit adds a page entry but forgets the base glossary entry). Instead of a non-null assertion, add an explicit guard with a clear error (or filter out missing entries) so the page fails loudly and diagnostically during dev/build rather than crashing unexpectedly.

clarkemoyer:

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: `aria-expanded` is being set to string values ('true'/'false'). Elsewhere in the codebase this attribute is passed a boolean (e.g. `aria-expanded={isOpen}`), and React’s ARIA typings also expect a boolean here. Prefer `aria-expanded={open}` to keep semantics consistent and avoid accidental string/boolean mismatches.
```suggestion
        aria-expanded={open}
```
2. `src/components/glossary-term/index.tsx`: The helper is named `closeTooltip`, but this component is explicitly rendering a `role="dialog"` popover (and the file/docs call it a popover/dialog). Renaming this helper (and related naming like `popoverId` vs `closeTooltip`) to consistently use “popover”/“dialog” would reduce confusion when maintaining the component.
```suggestion
  const closePopover = useCallback(
```

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: PR description says the inline term popover uses `role="tooltip"`/`aria-describedby`, but this component intentionally renders the popover as `role="dialog"` (interactive) and doesn’t set `aria-describedby` on the trigger. Please update the PR description (and any related docs) to match the implemented ARIA model, or adjust the implementation to match the stated behavior.
2. `src/components/glossary-term/index.tsx`: The trigger button isn’t programmatically associated with the popover content (no `aria-describedby` pointing to the definition), and the dialog itself has no `aria-describedby` for its body text. Because focus stays on the trigger when opening on focus/hover, many screen readers won’t announce the definition. Consider adding an id to the definition text and wiring `aria-describedby` (on the dialog, and/or conditionally on the trigger when open) so the definition is read reliably.
3. `tests/glossary-term.spec.ts`: This locator grabs the first `role="dialog"` on the page. The site already has other dialog-based UI (e.g., cookie preferences, mobile nav), so this test can become flaky if another dialog is present. Prefer scoping the dialog query (e.g., within the Term container) or using an accessible name filter (e.g., the entry title) to ensure you’re asserting against the glossary popover specifically.
4. `tests/glossary-term.spec.ts`: The outside-click portion relies on hard-coded mouse coordinates (5, 500). This is brittle across viewport sizes, responsive layout shifts, or content changes. Prefer clicking a stable outside element (e.g., page header/footer, a specific test id, or `page.locator('body')` with a safe position derived from the trigger’s bounding box).
```suggestion
    // Click somewhere outside the component using a stable element and a
    // layout-aware position. The popover opens ABOVE the trigger
    // (`bottom-full left-0`), so clicking below the trigger remains outside
    // both the trigger and the popover without relying on hard-coded
    // viewport coordinates.
    const triggerBox = await triggerButton.boundingBox()
    const body = page.locator('body')
    const bodyBox = await body.boundingBox()

    expect(triggerBox).not.toBeNull()
    expect(bodyBox).not.toBeNull()

    const clickX = Math.max(1, Math.min(10, bodyBox!.width - 1))
    const clickY = Math.min(
      bodyBox!.height - 1,
      Math.max(triggerBox!.y + triggerBox!.height + 24 - bodyBox!.y, 1),
    )

    await body.click({
      position: {
        x: clickX,
        y: clickY,
      },
    })
```

clarkemoyer:

clarkemoyer:

1. `src/components/glossary-term/index.tsx`: The trigger button never closes the popover on repeated activation (onClick always sets open=true). For ARIA-expanded disclosure/popup controls, users generally expect the same control to toggle (open/close). Consider toggling open state on click/keyboard activation when already open, while keeping hover/focus behavior unchanged.
2. `src/components/glossary-term/index.tsx`: PR description/issue text says the popover uses role="tooltip", but the implementation uses role="dialog" (and aria-haspopup="dialog"). If role="dialog" is the intended a11y approach (because it contains a focusable link), please update the PR description (and/or issue acceptance criteria) to avoid documenting the wrong ARIA pattern.
3. `src/app/results/reliability/page.tsx`: PR description/issue closure says this closes #1780, but the acceptance criteria for #1780 calls for wrapping terms on additional pages (factor-analysis/validation and CRP counterparts). This PR only wires Cronbach’s alpha on the two reliability pages; consider either expanding the wiring to meet the issue’s minimum page list or adjusting the PR/issue linkage so #1780 isn’t closed prematurely.

clarkemoyer:

|---|---|---|---|---|
| 1 | 7 | glossary-terms.ts, index.tsx | 🔧 Fixed | 2026-04-23 15:09 |
| 2 | 5 | index.tsx, glossary-terms.ts, page.tsx | 🔧 Fixed | 2026-04-23 15:32 |
| 3 | 4 | index.tsx, page.tsx | 🔧 Fixed | 2026-04-23 16:19 |
| 4 | 1 | page.tsx | 🔧 Fixed | 2026-04-23 16:56 |
| 5 | 2 | index.tsx | 🔧 Fixed | 2026-04-23 17:22 |
| 6 | 4 | index.tsx, glossary-term.spec.ts | 🔧 Fixed | 2026-04-23 17:47 |
| 7 | 1 | index.tsx | 🔧 Fixed | 2026-04-23 18:07 |
| 8 | 3 | index.tsx, page.tsx | 🔧 Fixed | 2026-04-23 19:09 |
| 9 | 0 | — | ✅ Clean | 2026-04-23 19:31 |

### Round 1 Comments (7)

- `src/data/glossary-terms.ts` — The `as const` assertion is effectively neutralized here because the array is already type-annotated as `readonly Glossa
- `src/data/glossary-terms.ts` — The `as const` assertion is effectively neutralized here because the array is already type-annotated as `readonly Glossa
- `src/components/glossary-term/index.tsx` — Using `React.ReactNode` requires the `React` namespace type to be available in this module. To make this robust across T
- `src/components/glossary-term/index.tsx` — Using `React.ReactNode` requires the `React` namespace type to be available in this module. To make this robust across T
- `src/components/glossary-term/index.tsx` — The PR description says the tooltip closes on “click-outside” for tap as well, but the outside handler is wired to `mous
- `src/components/glossary-term/index.tsx` — `buttonRef` is currently not referenced anywhere besides being passed to the `ref` prop. If there’s no planned focus/mea
- `src/components/glossary-term/index.tsx` — The click-outside-to-close behavior is newly introduced but not directly covered by tests. Add a unit test that opens th

### Round 2 Comments (5)

- `src/components/glossary-term/index.tsx` — Escape handling assumes the trigger still has focus, but focus can move into the tooltip (the “See full entry” link is f
- `src/components/glossary-term/index.tsx` — The popover includes a focusable Next.js <Link> inside an element with role="tooltip". Per ARIA tooltip patterns, toolti
- `src/data/glossary-terms.ts` — This PR claims glossary-terms.ts is the single source of truth for glossary copy, but /src/app/results/glossary/page.tsx
- `src/components/glossary-term/index.tsx` — There is already a tooltip implementation in src/components/ui/tooltip.tsx; this introduces a second tooltip/popover pat
- `src/app/results/reliability/page.tsx` — This change adds new interactive hover/focus/tap behavior on a production results page, but there is no Playwright cover

### Round 3 Comments (4)

- `src/components/glossary-term/index.tsx` — The trigger button isn’t programmatically associated with the popover content. The dialog element has an `id={popoverId}
- `src/components/glossary-term/index.tsx` — PR description says the popover uses `role="tooltip"` and `aria-describedby`, but the implementation explicitly uses `ro
- `src/app/results/glossary/page.tsx` — `const base = _termBase.get(extra.id)!` will throw at runtime if a PAGE_EXTRAS id is missing from `glossaryTerms` (or if
- `src/components/glossary-term/index.tsx` — The popover is closed immediately on `onMouseLeave` for the trigger button. Because the popover is positioned with a `mb

### Round 4 Comments (1)

- `src/app/results/glossary/page.tsx` — `_termBase.get(extra.id)!` will throw at runtime if `PAGE_EXTRAS` ever contains an id that’s missing from `glossaryTerms

### Round 5 Comments (2)

- `src/components/glossary-term/index.tsx` — `aria-expanded` is being set to string values ('true'/'false'). Elsewhere in the codebase this attribute is passed a boo
- `src/components/glossary-term/index.tsx` — The helper is named `closeTooltip`, but this component is explicitly rendering a `role="dialog"` popover (and the file/d

### Round 6 Comments (4)

- `src/components/glossary-term/index.tsx` — PR description says the inline term popover uses `role="tooltip"`/`aria-describedby`, but this component intentionally r
- `src/components/glossary-term/index.tsx` — The trigger button isn’t programmatically associated with the popover content (no `aria-describedby` pointing to the def
- `tests/glossary-term.spec.ts` — This locator grabs the first `role="dialog"` on the page. The site already has other dialog-based UI (e.g., cookie prefe
- `tests/glossary-term.spec.ts` — The outside-click portion relies on hard-coded mouse coordinates (5, 500). This is brittle across viewport sizes, respon

### Round 7 Comments (1)

- `src/components/glossary-term/index.tsx` — When focus moves from the popover’s internal link to an element outside the component (e.g., Tab past “See full entry”),

### Round 8 Comments (3)

- `src/components/glossary-term/index.tsx` — The trigger button never closes the popover on repeated activation (onClick always sets open=true). For ARIA-expanded di
- `src/components/glossary-term/index.tsx` — PR description/issue text says the popover uses role="tooltip", but the implementation uses role="dialog" (and aria-hasp
- `src/app/results/reliability/page.tsx` — PR description/issue closure says this closes #1780, but the acceptance criteria for #1780 calls for wrapping terms on a

### Result

✅ All automated review comments addressed in 2 round(s). Ready for final human review.


**Copilot Review Cycle:** ✅ Passed after 2 round(s) - review clean, all CI checks green. Ready to merge.

Issue #1787, opened 2026-04-23
fix(mind-maps): cut pages render cropped viewBox that shrinks content vs showing full map with initial focus


clarkemoyer:
## Problem

The 7 Mind Map cut pages (`/making-of-tabs/mind-maps/{business-management-models, it-management-models, enterprise-it-architecture, project-program-risk-management, standards-regulations, tabs-project-operations, culminating-research-project}`) each serve a **viewBox-cropped copy** of the full mind map SVG at `/Svgs/mind-maps/{slug}.svg`.

The crop works — but it chops the SVG down so:

- Users can see only the cropped region; panning beyond the crop produces blank area, not adjacent content
- The rendered pixel dimensions are smaller (e.g. 5103 × 2620 instead of 10207 × 6731) so at the same fit-to-wrapper scale the labels are smaller and less readable
- Users cannot scroll to context around the cut

## Proposed fix

Keep the 7 cut pages but change their `MindMapViewer` invocation from a separate cropped SVG to the **full map** with an `initialFocus` prop that pans/zooms the viewer to the target region on load.

- `src/components/mind-maps/mind-map-viewer.tsx`: accept `initialFocus?: { x, y, w, h }` (in SVG coordinates) and, on mount, compute a transform that centers that region at a fit-zoom
- `src/app/making-of-tabs/mind-maps/<slug>/page.tsx` (all 7): change `src` to `/Svgs/mind-maps/full-mind-map.svg` and pass the region's viewBox values as `initialFocus`
- Delete the 7 standalone cropped SVG files in `public/Svgs/mind-maps/{slug}.svg` — full map becomes the single source

## Why

- Lets users pan/zoom past the initial region to see adjacent context (e.g. on the "Business Management Models" cut, they can scroll to "IT & IT Management Models" without navigating to a different page)
- The fitted scale for the full map is still the same; content legibility unchanged at the focused region
- Removes ~8 MB of near-duplicate SVG files
- Pairs naturally with #1786 (full-screen + granular zoom) since the full map rewards deeper exploration

## Acceptance

- [ ] Each cut page loads with its region centered and at a readable zoom
- [ ] User can pan/zoom beyond the initial region to see the rest of the map
- [ ] Removing the 7 cropped SVGs doesn't break any route or sitemap link
- [ ] Playwright: verify cut pages still render the viewer and `naturalWidth` matches full map (10207)

## Related

- #1786 full-screen + granular zoom controls
[...]

PR #1789, opened 2026-04-23
feat(mind-maps): fullscreen + keyboard + zoom slider, route cuts via initialFocus

### LINKED ISSUE #1786: feat(mind-maps): full-screen pan/zoom/scroll viewer with granular controls
## User-facing problem

The current Mind Map viewer at `/making-of-tabs/mind-maps/*` has a coarse 3-button toolbar (+ / − / Reset) with big zoom steps. Users cannot:

- Enter fullscreen to get the maximum readable view
- Pan smoothly (click-drag works but keyboard doesn't)
- Zoom with fine-grained control (small steps, mouse-wheel sensitivity)
- Scroll naturally

## What

Upgrade `src/components/mind-maps/mind-map-viewer.tsx`:

1. **Fullscreen toggle** — Fullscreen API, with `F` keyboard shortcut and a toolbar button
2. **Keyboard controls** — arrow keys pan, `+` / `-` zoom, `0` reset to fit, `F` fullscreen, `Esc` exit fullscreen
3. **Mouse-wheel zoom** — continuous, small step (0.1 or smaller), centered on cursor
4. **Zoom slider** — toolbar gets a range slider so users can drag to any scale between `minScale` and `maxScale`
5. **Smaller button steps** — current `step: 0.1` via wheel is OK; add smaller click-button steps (0.05) and shift-click for larger (0.25)
6. **Pan indicator / minimap** (stretch) — small preview box in the corner showing current viewport position on the full map

## Why

Mind maps are 10207 × 6731 pixels. At wrapper size of ~1200 × 720, the fitted scale is ~0.11. To read any individual node label you need to zoom to roughly 0.5–1.0. The current +/− button jumps in 0.1 increments, so reaching a useful zoom takes 10+ clicks. Fullscreen + wheel zoom + fine steps should make exploration feel natural.

## Test plan

- [ ] Keyboard: arrows pan, +/- zoom, 0 reset, F fullscreen, Esc exits
- [ ] Mouse wheel zooms smoothly toward cursor position
- [ ] Slider drags in real-time
- [ ] Fullscreen works on desktop (Chrome/Firefox/Safari)
- [ ] Accessibility: all new controls have aria-labels and keyboard accessibility
- [ ] No regression on existing pan-drag behavior

## Not in scope

- Minimap (stretch; may defer)
- Touch-gesture tuning (pinch already works via library; don't rewrite)
- Per-map initial zoom/pan presets (covered by separate viewBox-restore issue)

### LINKED ISSUE #1788: fix(nav): Mind Maps not surfaced in sidebar on Home / from outside Making-of-TABS
## Problem

The left Sidebar hides the **Mind Maps** group until the user is already on a `/making-of-tabs/*` page:

1. From Home (`/`), the sidebar's active section is `home` which has no groups. The user has to click **Making of TABS** in the section picker to see any group, then expand Mind Maps.
2. From any non-Making-of-TABS section (Results, Models, Teaching, etc.) the Mind Maps group isn't visible at all.
3. Even when on `/making-of-tabs/` (the section landing page, not the Mind Maps sub-route), the Mind Maps accordion group may not be auto-expanded — the user has to click to open.

Result: Mind Maps is a hidden two-click-away destination.

## Proposed fix

Several reasonable options; any one (or combination) works:

**Option A — Auto-expand "Mind Maps" when a user is on any `/making-of-tabs/*` page.**
The Sidebar already auto-picks the section from pathname; extend the same logic to auto-open the matching group. If `pathname.startsWith('/making-of-tabs/mind-maps')` then set `openGroup = 'Mind Maps'`.

**Option B — Promote Mind Maps to a top-level section.**
Its own icon in the section picker row, alongside Home / Survey / Results / etc. This makes it a first-class destination, not hidden under Making-of-TABS.

**Option C — Surface a "Featured" or "Visualizations" link from Home.**
The `home` section currently has `groups: []`. Add a small featured-links group that always points into the mind-maps gallery.

## Why

- The mind maps are genuinely novel content that users are likely to want to return to; hiding them two clicks deep buries them.
- If the site is aiming for "show the research visually" as a selling point, the nav should reflect that.

## Acceptance

- [ ] From Home, a user can reach the Mind Maps gallery in at most one click via the sidebar
- [ ] Path-based auto-open of the Mind Maps accordion group is respected on all `/making-of-tabs/mind-maps/*` routes
- [ ] jest-axe passes on the updated sidebar
- [ ] No regression to the section picker / keyboard-navigation behavior

## Preferred option

Option A is the minimal / least-invasive fix. Option B is the best UX if we're willing to tweak the top-level nav. Recommend starting with A + a cross-section "Featured" link on Home, and keeping Option B on the table.

clarkemoyer:
## Summary

Rewrites the Mind Map viewer so users can actually use it at the high zoom levels the 10207x6731 Lucidspark SVG demands, and stops shipping seven separately-cropped copies of the same artboard.

Closes #1786 and #1787.

## Viewer changes (`MindMapViewer`)

- **Fullscreen toggle** via Fullscreen API. Button label + `aria-pressed` track state via `fullscreenchange`.
- **Zoom slider** (`<input type="range">`) alongside the existing zoom in / zoom out / reset buttons, labelled "Zoom level".
- **Keyboard controls** (viewer region is `tabIndex={0}`):
  - Arrow keys pan 80px
  - `+` / `-` zoom by 0.15
  - `0` reset to fit / initial focus
  - `F` toggle fullscreen
- **Wider zoom range**: `MIN_SCALE` 0.02, `MAX_SCALE` 8 (was 0.1 / 4). Lets the viewer frame the full 10207x6731 artboard and also zoom deep into leaf nodes.
- **`initialFocus` prop**: optional `{ x, y, w, h }` in SVG coordinates. When provided, the viewer centers that region on initial mount, resize, and reset.

## Route changes (cuts now sit over the full SVG)

All seven branch-cut pages now render `/Svgs/mind-maps/full-mind-map.svg` with an `initialFocus` region instead of a dedicated cropped SVG:

| Cut | initialFocus (x, y, w, h) |
| --- | --- |
| business-management-models | 2900, 4500, 3700, 2200 |
| culminating-research-project | 5300, 700, 3700, 1100 |
| enterprise-it-architecture | 0, 4500, 2900, 2200 |
| it-management-models | 0, 2600, 3700, 1900 |
| project-program-risk-management | 6600, 4500, 3600, 2200 |
| standards-regulations | 1800, 2600, 1900, 1900 |
| tabs-project-operations | 7400, 0, 2800, 1600 |

Each page's prose updated to describe the new behavior. Mind Maps index drops the stale "rough viewBox placeholder" caveat. The seven cropped SVG files are removed from `public/Svgs/mind-maps/` (full map remains).

## Test plan

- [x] `npm run lint` clean
- [x] `npm test` — 564 tests pass (55 suites), including existing `MindMapViewer` toolbar + a11y checks
- [x] `npm run build` — static export succeeds, 268 pages generated
- [ ] Manual QA: load each cut page, verify branch sits centered with room to pan to neighbours
- [ ] Manual QA: fullscreen + keyboard shortcuts behave per docs
- [ ] `npm run test:e2e`

Does not close #1788 (nav visibility). That will follow in a separate PR.


clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: `onTransform={(_, state) => setCurrentScale(state.scale)}` runs on every pan/drag frame as well as zoom, which can trigger frequent React re-renders while users pan the map. Since the toolbar only needs the scale, consider updating state only when `state.scale` changes (e.g., compare with previous value), or use a throttled/RAF update to reduce render pressure during panning.
```suggestion
            onTransform={(_, state) =>
              setCurrentScale((previousScale) =>
                previousScale === state.scale ? previousScale : state.scale
              )
            }
```
2. `src/components/mind-maps/mind-map-viewer.tsx`: `focusRegion()` assumes `region.w` and `region.h` are positive and finite; if a caller passes `0`, negative, or `NaN`, `focusScaleFor()` can yield `Infinity`/`NaN` and the viewer will jump to an invalid transform. Add input validation/guards (e.g., return early unless `w > 0 && h > 0` and all values are finite) before computing the scale and center.
```suggestion

      const { x: regionX, y: regionY, w: regionW, h: regionH } = region
      if (
        !Number.isFinite(regionX) ||
        !Number.isFinite(regionY) ||
        !Number.isFinite(regionW) ||
        !Number.isFinite(regionH) ||
        regionW <= 0 ||
        regionH <= 0
      ) {
        return
      }

      const scale = clamp(focusScaleFor(w, h, regionW, regionH), MIN_SCALE, MAX_SCALE)
      const regionCx = regionX + regionW / 2
      const regionCy = regionY + regionH / 2
      const x = w / 2 - regionCx * scale
      const y = h / 2 - regionCy * scale
      if (!Number.isFinite(scale) || !Number.isFinite(x) || !Number.isFinite(y)) return
```
3. `src/components/mind-maps/mind-map-viewer.tsx`: `fitToWrapper()` computes a fit scale but does not clamp it to `MIN_SCALE`/`MAX_SCALE`, even though `TransformWrapper` enforces those bounds. On very small viewports the fitted scale can fall below `MIN_SCALE`, causing `currentScale` (and the slider) to disagree with the transform state. Clamp the computed `scale` before calling `setTransform()` / `setCurrentScale()`.
4. `src/components/mind-maps/mind-map-viewer.tsx`: The container-level `onKeyDown` handler will also receive bubbled key events from the zoom slider (`<input type="range">`). Because arrow keys are handled with `preventDefault()`, keyboard users won’t be able to adjust the slider with ArrowLeft/ArrowRight/ArrowUp/ArrowDown while it’s focused. Consider only handling these shortcuts when `e.currentTarget === e.target` (container itself has focus), or early-return when the event target is an interactive control (e.g., input/range/buttons).

clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: `toggleFullscreen` calls `el.requestFullscreen()` / `document.exitFullscreen()` without handling promise rejections, and without guarding for browsers where the Fullscreen API isn’t available on arbitrary elements (common on iOS Safari). This can throw or produce unhandled promise rejections when fullscreen is denied/unsupported. Consider feature-detecting (`el.requestFullscreen` / `document.exitFullscreen`) and handling the returned promises (e.g., catch and/or fall back to no-op) so the viewer doesn’t break in unsupported environments.
```suggestion

    if (document.fullscreenElement === el) {
      if (typeof document.exitFullscreen !== 'function') return
      void document.exitFullscreen().catch(() => {})
      return
    }

    if (typeof el.requestFullscreen !== 'function') return
    void el.requestFullscreen().catch(() => {})
```
2. `src/components/mind-maps/mind-map-viewer.tsx`: The fullscreen/keyboard handler is attached to a `div` that’s inserted into the tab order via `tabIndex={0}`, but it doesn’t have an explicit accessible name and the comment mentions `outline-none` even though no focus styling class is applied. This makes it hard for keyboard/screen-reader users to understand what they’ve focused. Consider either moving `tabIndex`/`onKeyDown` onto the already-labelled `<section>` or adding an `aria-label`/`aria-labelledby` plus an intentional `focus-visible` style on this focusable container (and updating/removing the misleading comment).
3. `src/components/mind-maps/mind-map-viewer.tsx`: `MindMapViewer` adds new user-facing behaviors (zoom slider, fullscreen toggle, keyboard shortcuts, `initialFocus` reset behavior), but the existing unit tests only assert the presence of the toolbar buttons and run `jest-axe`. Adding focused tests for the new controls/shortcuts (e.g., slider renders with the expected label and updates zoom, fullscreen button renders with correct pressed state, keydown pans/zooms when the container is focused) would help prevent regressions.

clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: The toolbar’s accessible name is still “Mind map zoom controls”, but it now contains non-zoom actions (Reset + Fullscreen). This is misleading for screen reader users; consider updating the aria-label to something like “Mind map controls” or “Mind map zoom and fullscreen controls”.
```suggestion
            aria-label="Mind map controls"
```
2. `src/components/mind-maps/mind-map-viewer.tsx`: `initialFocus` introduces new framing behavior (initial mount, resize, reset, double-click) but the unit tests don’t cover it. Adding a test that renders with `initialFocus` and asserts the viewer resets to that focus (e.g., by mocking react-zoom-pan-pinch’s ref/setTransform) would protect against regressions in the focusRegion math and the reset/double-click branching.

clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: `onKeyDown` only exempts form controls by tagName. When focus is on the "Open raw SVG" link (or any other interactive element like an `<a>`), arrow/+/−/0/F will still be intercepted and `preventDefault()` will run, which can break expected keyboard interaction. Consider expanding the guard to include anchors (and/or checking `closest('a,button,input,select,textarea,[contenteditable]')`) so shortcuts only apply when the viewer container itself is focused.
```suggestion
      // Don't steal arrow/key events from interactive controls (e.g. the zoom slider)
      // or links/contenteditable elements inside the viewer.
      const target = e.target
      if (
        target instanceof HTMLElement &&
        target.closest(
          'a,button,input,select,textarea,[contenteditable],[contenteditable="true"],[contenteditable=""]'
        )
      ) {
        return
      }
```
2. `__tests__/components/mind-maps/mind-map-viewer.test.tsx`: This test attempts to spy on `preventDefault()` by passing a `preventDefault` function in the `fireEvent.keyDown` init object, but `preventDefault` is a method on the Event prototype and isn’t reliably overridden this way. As written, the assertion can pass even if the handler calls `e.preventDefault()`. To make this meaningful, create the KeyboardEvent first (e.g. via `createEvent.keyDown`), `jest.spyOn(event, 'preventDefault')`, then dispatch it, or assert `event.defaultPrevented` is false.

clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: Keyboard shortcuts only fire when the `<section>` itself (or a non-interactive descendant) has focus. Clicking/dragging the image likely won’t move focus to the section, so users may see arrow/F/+/- shortcuts “not working” after interacting with the map. Consider focusing the section/container on pointer-down (or add explicit instructions like “When the viewer is focused…” in the help text).
2. `__tests__/components/mind-maps/mind-map-viewer.test.tsx`: New keyboard behavior is only partially covered: there’s a test that shortcuts don’t preventDefault when the slider has focus, but no test that the viewer shortcut handler *does* preventDefault when the viewer itself is focused (and thus would pan/zoom). Adding a small test that dispatches a keydown on the section and asserts `defaultPrevented === true` would guard against regressions.
```suggestion

  it('keyboard shortcuts preventDefault when the viewer itself has focus', () => {
    const { container } = render(<MindMapViewer {...defaultProps} />)
    const viewer = container.querySelector('section')
    expect(viewer).not.toBeNull()

    // Use a real KeyboardEvent and check defaultPrevented, because fireEvent
    // init-object overrides of `preventDefault` are not reliable.
    const event = new KeyboardEvent('keydown', {
      key: 'ArrowRight',
      bubbles: true,
      cancelable: true,
    })

    viewer?.dispatchEvent(event)
    expect(event.defaultPrevented).toBe(true)
  })
```

clarkemoyer:

1. `src/components/mind-maps/mind-map-viewer.tsx`: If `initialFocus` is provided but invalid (NaN/0/negative dims), `focusRegion` returns early and the viewer never falls back to fitting the full image—leaving the transform at the library defaults (often top-left at `initialScale`). Consider having `focusRegion` return a boolean (success/failure) and falling back to `fitToWrapper()` when it fails, or validating `initialFocus` once and treating invalid input as `undefined`.
2. `src/components/mind-maps/mind-map-viewer.tsx`: `currentScale` is initialized to `MIN_SCALE` (0.02) but `TransformWrapper` starts at `initialScale={0.08}` and the component doesn't call `fitToWrapper`/`focusRegion` until after the image dimensions are known. This can make the slider briefly show 2% while the actual transform is 8% (or another value) on first paint. Consider initializing `currentScale` to the same `initialScale`, or hiding/disabled the slider until the initial framing runs.

clarkemoyer:

Issue #16, opened 2026-01-15
Create Prolific workflows and supported functions with the API


clarkemoyer:
We are using https://app.prolific.com/ to capture valid survey results. We need to set up the collection via API. A new environment "prolific-prod" has been created in this repo and a new token under it called "TABS_PROLIFIC_TOKEN" has been made. Create a workflow that queries this API and gains information on the survey that it is able to use. Also note any other features of the API. Update documentation to support this new capability of the repo.

PR #1837, opened 2026-05-01
Statistical extensions to TABS validation pipeline - 28 new analyses (closes #1836)

### LINKED ISSUE #1836: Statistical extensions - add advanced psychometric analyses to validation pipeline
## Background

A statistical-rigor review of the TABS validation pipeline identified gaps in the factor-analytic and inferential coverage. The 50-reviewer rounds and committee feedback raised specific questions about exact factor-analysis steps performed, single-factor vs multi-factor structure, item-level convergent and discriminant validity, mediation, measurement invariance, ordinal estimator choice, and replicability.

This issue captures the comprehensive statistical extension that adds the missing analyses to the production pipeline so they regenerate automatically for every daily run on both the live ongoing dataset (`live-validation.json`) and the frozen CRP-200 dataset (`crp-validation.json`). All extensions are additive to the existing JSON schema and run unconditionally - no opt-in flag, no separate analysis path.

### Multi-spec confirmatory factor analysis on Barriers
1. **Barriers 3-factor canonical CFA (F1a / F1b / F2)** - tests the decomposition reported in the CRP body as a measurement model. Best fit indices of any Barriers spec at N=200 (CFI 0.886, RMSEA 0.069 with ML; CFI 0.990, RMSEA 0.031 with DWLS).
2. **Barriers 2-factor EFA-derived CFA** - matches the parallel-analysis result (F1=14, F2=4 items).
3. **Formal CFA model comparison** - chi-squared difference, delta-AIC, delta-BIC across 1F / 2F / 3F / 4F barrier specs. Confirms 3F is significantly better than 2F (p<.001), which is significantly better than 1F (p<.001).
4. **Joint three-construct CFA** - Barriers + Readiness + Maturity in one measurement model with free latent correlations. Latent r(R, M) = 0.80 (matches the existing HTMT 0.804).
5. **Second-order Barriers CFA** - F1a / F1b / F2 -> Barriers higher-order. DWLS CFI 0.990, RMSEA 0.031. Justifies treating Barriers as a single construct in regressions.
6. **Confirmatory bifactor on Barriers** - G + 3 specific factors. ECV(G) 0.65, omega-h(G) 0.79; specific factors are weak. Barriers is essentially unidimensional.

### Estimator-correct CFA for ordinal data
7. **DWLS ordinal CFA** - Mardia's test rejects multivariate normality for all three scales; ML standard errors are biased. DWLS (the WLSMV-equivalent for `semopy`) is the methodologically correct estimator. Re-runs every CFA model with DWLS.
8. **Mardia's multivariate normality test** - formal justification for DWLS over ML.
9. **Mahalanobis multivariate outliers** - 0-1 outliers per scale at chi-squared p<.001; sample is clean.

### Item-level psychometrics
10. **Hair-style item-level convergent / discriminant tests** - primary loading >= 0.50 (acceptable) + cross-loading < 0.30 OR primary minus secondary >= 0.20.
11. **IRT graded response model (Samejima GRM)** - per-item discrimination + thresholds + ceiling-effect flags. B13 / B14 (Cybersecurity / Privacy) flagged as ceiling-effect items.
12. **Differential Item Functioning (DIF)** - per-item IRT discrimination by SMB vs Enterprise; flag |delta| > 0.5.
13. **Alpha-if-deleted summary** - per-construct count of items raising alpha if removed.
14. **Per-item Cohen's d (SMB vs Enterprise)** - identifies the largest item-level group differences (M5 Tech Risk d=0.46; B14 Privacy d=0.39).

### Subgroup analyses
15. **HTMT bootstrap CI + Fornell-Larcker on F1a / F1b / F2** - treats subgroups as 3 constructs. F1a vs F1b HTMT 0.88 (FAIL), confirming F1a / F1b are theoretical labels for one empirical factor.
16. **Per-subgroup standalone validation** - each subgroup's items run as a 1F CFA in isolation. F1a + F1b each pass alpha + CFI + RMSEA; F2 has perfect factor structure but alpha 0.67.
17. **Multi-group SEM (3F Barriers, SMB vs Enterprise)** - the structural finding: Enterprises see Barriers as more unified than SMBs do.
18. **Reliability by demographic group** - alpha per construct for SMB / Enterprise / For-Profit / Non-Profit subsamples.

### Inferential extensions
19. **Mediation analysis (Barriers -> Readiness -> Maturity)** - bootstrapped indirect effect (n=2000). Frozen N=200: indirect = -0.305 (p<.001), direct = -0.055 (n.s.). Readiness FULLY MEDIATES Barriers -> Maturity.
20. **Per-factor regressions** - sub-factor decomposition vs total Barriers. Reveals sign-reversal masked by aggregating: F1a / F1b NEGATIVELY predict R / M; F2 (External & Compliance) POSITIVELY predicts both.
21. **Standardized regression coefficients + VIF** - effect-size comparisons across F1a / F1b / F2; multicollinearity check (all VIF < 2.1).
22. **Equivalence testing (TOST, +/- 0.30 SD)** - converts non-significant t-tests into formal equivalence statements. Barriers means are statistically equivalent SMB vs Enterprise; R / M are not.
23. **Power analysis** - smallest detectable Cohen's d at current N (0.40 at N=200; 0.26 at N=500).

### Robustness + replication
24. **Bootstrap CIs on Cronbach alpha** - 1000-iter bootstrap percentile CIs supplementing the Feldt CI.
25. **Split-sample cross-validation** - random 50/50 split, refit 3F canonical, compute Tucker congruence on standardized loadings (frozen 0.984; live 0.987).
26. **ESEM (3-factor with oblimin rotation)** - allows free cross-loadings. Suggests B18 (Vendor / Partner Difficulty) reassign from F2 to F1b in V3.
27. **HTMT bootstrap CI** (already in JSON; kept for completeness; now extended to subgroups).
28. **Construct stability across sample tiers** - max range of construct mean across 5 tiers; supports CRP §4.4.4 sensitivity framing.

## Where each output surfaces on the website (no new URLs)

| Output | CRP page | Live page | Section to extend |
|---|---|---|---|
| Multi-spec CFA + comparison + DWLS | `/results/crp-2026/validation` | `/results/validation` | Section 3: CFA |
| Joint 3-construct CFA | `/results/crp-2026/validation` | `/results/validation` | Section 3: CFA |
| Bifactor (R+M; Barriers) | `/results/crp-2026/validation` | `/results/validation` | New "Bifactor" section |
| Mardia + Mahalanobis | `/results/crp-2026/validation` | `/results/validation` | New "Assumption Checks" section |
| Item-level Hair + IRT + DIF | `/results/crp-2026/validation` | `/results/validation` | Section 6: Item Diagnostics |
| Subgroup HTMT/FL + standalone validation + 3F canonical CFA | `/results/crp-2026/factor-analysis` | `/results/factor-analysis` | Level 3: 3-Group Decomposition |
| Cross-validation + Tucker congruence | `/results/crp-2026/factor-analysis` | `/results/factor-analysis` | Level 3 (replication footer) |
| Alpha-if-deleted + bootstrap CI + by-demographic alpha | `/results/crp-2026/reliability` | `/results/reliability` | After "Cronbach's Alpha by Construct" |
| Mediation + per-factor regression + standardized betas + VIF + TOST + power | `/results/crp-2026/findings` | `/results/findings` | New "Inferential Extensions" section |
| ESEM | `/results/crp-2026/factor-analysis` | `/results/factor-analysis` | New "ESEM Sensitivity" section |

The CRP-2026 and Live tracks share React components (`FactorAnalysisContent`, `ValidationPageContent`) with a `variant` prop, so component edits light up both pages.

## Acceptance criteria

- [x] All 28 new top-level JSON keys are populated for both `crp-validation.json` and `live-validation.json` on every daily pipeline run.
- [x] No CLI flag, no opt-in, no separate analysis path - the extensions run unconditionally.
- [x] Existing JSON schema keys are unchanged.
- [x] Reproducibility verified: frozen CRP-200 numbers cross-checked between an ad-hoc Python run and the production scripts.
- [ ] Follow-up PR (issue #1839) edits the React components to surface the new keys on existing pages.

## Related

- PR #1837: implementation
- Issue #1839: website page updates + permanent data flow diagram (next step after this PR merges)

clarkemoyer:
Closes #1836

Adds 28 advanced psychometric and inferential analyses to the TABS validation pipeline. All run unconditionally on every daily invocation for both the live ongoing dataset and the frozen CRP-200 dataset. Additive to the existing JSON schema; existing keys are unchanged.

## Pre-merge verification + methodological additions (2026-05-01 final round)

The latest two commits add a methodological cross-check layer and resolve the
remaining Copilot review items, after a full R-vs-Python parity test against
the frozen CRP-200 dataset using locally-installed R 4.6.0 + psych 2.6.3 +
lavaan 0.6.21 + semTools 0.5.8 + MVN 6.3.

### What's new in the JSON output (additive only - no breaking changes)

| Field | Where | What it is |
|---|---|---|
| `bifactor_barriers.omega_t` | per sample | McDonald omega-total (variance from general + group factors); the directly-comparable counterpart to R's `psych::omega(...)$omega.tot`. Reflects the 3-factor F1a/F1b/F2 structure rather than only the 1F-CFA-derived McDonald omega. CRP-200 value: 0.897 vs R's 0.896. |
| `htmt[*].htmt2` | per pair | HTMT2 (Roemer, Schuberth & Henseler 2021) - the geometric-mean variant that semTools 0.5.6+ uses by default. Reported as a robustness check alongside the canonical `htmt` (Henseler 2015). All three CRP-200 pairs stay below 0.85 under both variants. |

### Why both omega variants

For unidimensional scales (Readiness, Maturity per parallel analysis), the
1-factor McDonald omega already in `mcdonalds_omega` is the appropriate
statistic (Cho 2016; Revelle & Condon 2019). For Barriers, where the
canonical 3-factor structure is the load-bearing claim of the dissertation,
bifactor `omega_t` is the more theoretically appropriate reliability
(Reise 2012; Rodriguez, Reise & Haviland 2016). The pipeline now reports
both so the paper can defend either choice.

### Why both HTMT variants

`htmt` (Henseler 2015 - arithmetic mean) is the version most-cited in
published research and the version against which the canonical 0.85 / 0.90
discriminant-validity thresholds were calibrated. `htmt2` (Roemer 2021 -
geometric mean) has better theoretical properties (consistent estimator
under tau-equivalence) and is what semTools defaults to in current versions.
Reporting both pre-empts a "which HTMT?" question without forcing a choice.

### Verification layers

1. **In-CI parity tests** (16/16 passing) -
   `scripts/analysis/tests/test_parity_to_published_formulas.py` validates
   every custom statistic against its published source-paper formula
   (Cronbach 1951, Fornell & Larcker 1981, Henseler 2015, Roemer 2021,
   Tucker 1951, Mardia 1970, Spearman 1910, Mahalanobis 1936, McDonald
   1999, Zinbarg et al. 2005). Includes the new HTMT2 formula check, the
   AM-GM ordering check (HTMT2 <= HTMT), the bifactor `omega_t >=
   omega_h_general` check, and a true split-half pipeline parity test
   (replacing the prior math-only identity check that didn't exercise
   pipeline code).

2. **Independent Python cross-check** - `pingouin` (Vallat 2018) is used
   in three tests as a third-party validation layer for Cronbach's alpha
   and multivariate normality.

3. **R cross-validation script** -
   `scripts/analysis/verify_against_R.R` computes every load-bearing
   statistic via `psych`, `lavaan`, `semTools`, and `MVN`, and writes
   results to JSON for direct diff against `src/data/crp-validation.json`.
   The script computes BOTH the 1F-omega and bifactor-omega in R, and BOTH
   HTMT (Henseler 2015) and HTMT2 (Roemer 2021), so each Python statistic
   has a directly-comparable R counterpart. Not a runtime dependency of
   the daily pipeline; the in-CI parity check is layer 1.

   Local verification result: every directly-comparable statistic agrees
   with the Python pipeline to <=0.014 absolute difference. Remaining
   small differences are listwise-vs-pairwise data scope (R does listwise
   on the joint 3-construct dataset; Python does pairwise on each
   construct-pair), not formula bugs.

### Review items addressed in this round

Three live Copilot review concerns from the 20:27 UTC review round, plus
two pre-existing bugs found while attempting to run the unified pipeline
end-to-end on Q4_OrgSize-present data:

| Item | Resolution |
|---|---|
| Duplicate `output['measurement_invariance'] = ...` (validation.py + unified) | Removed |
| `measurement_invariance` undefined in Q4_OrgSize-absent else branch (validation.py) | Initialized to `{'error': ...}` |
| `test_spearman_brown_split_half` claimed to exercise pipeline code but only checked the math identity | Rewritten to call `v.split_half_reliability()` on synthetic data and assert agreement with the manually-computed Spearman-Brown result |
| `mardia_multivariate_normality` / `mahalanobis_outliers` referenced `stats.chi2/norm` but scipy.stats was imported as `sp` (NameError) | Switched to `sp.chi2 / sp.norm` consistent with the rest of the file |
| `bootstrap_alpha_ci` and `reliability_by_demo` called undefined `cronbach_alpha` (function on this script is `cronbach_alpha_pd`) | Fixed both call sites |

### Cross-platform verification artifacts (Minitab + SPSS)

The descriptive + reliability layer of the pipeline can now be reproduced
end-to-end in Minitab and SPSS without writing any code or hunting through
GUI dialogs. New `scripts/minitab/` and `scripts/spss/` folders, plus three
per-tool shareable zips in `scripts/cross-platform-verification/`.

| Tool | Folder | One-time setup | Run-time clicks |
|---|---|---|---|
| **SPSS Statistics 31.0** (Base + Regression + Bootstrapping + Missing Values + Advanced Statistics) | `scripts/spss/` | Open `tabs_v2_validation.sps` | 2 clicks: open syntax, Run -> All |
| **Minitab** | `scripts/minitab/` | Open the .csv worksheet | 3 clicks: open worksheet, Run an Exec, select .MTB |
| **Shareable zips** | `scripts/cross-platform-verification/tabs_v2_validation_{spss,minitab,python}.zip` | Unzip on colleague's machine | Same as above |

**License-aware**: the SPSS syntax was built against the user's actual
installed components (Statistics Base + Regression + Bootstrapping +
Missing Values + Advanced Statistics; no IBM SPSS Amos). It uses the
Bootstrapping module for proper bootstrap CIs around Cronbach's alpha
(1000 resamples, matches Python's `bootstrap_alpha_ci` default) and the
Missing Values module for Little's MCAR test (matches the data-quality audit).

**Honest scope statement** in both READMEs: CFA-derived statistics
(McDonald's omega from CFA, composite reliability with SEs, AVE, HTMT,
HTMT2, bifactor decompositions, second-order CFA, multigroup CFA,
measurement invariance, ESEM, IRT GRM, Mardia normality, Tucker
congruence) cannot be computed in either Minitab or SPSS-without-Amos
regardless of macro. The recommendation is to cite the Python pipeline
(cross-validated against R via `scripts/analysis/verify_against_R.R`)
for those, and use SPSS / Minitab for the descriptive layer the macros
cover. Manual-from-EFA-loadings formulas for omega, CR, AVE, and Tucker
congruence are documented for users who want to do something in
SPSS/Minitab anyway.

**Verified parity**: Cronbach alpha for Barriers = 0.8728 in all four
platforms (Python `cronbach_alpha`, R `psych::alpha`, SPSS `RELIABILITY`,
Minitab `ItemAnal`); listwise N = 192 / 181 / 191 in all four. KMO =
0.851 / 0.927 / 0.912; Bartlett chi-squared = 1135.51 / 1278.99 / 650.13.

## Branch state (final)

- 33+ commits ahead of main, 0 behind, mergeable.
- Diff: 4 files changed.
   - `scripts/analysis/tabs_v2_validation.py` (~1500 lines added)
   - `scripts/analysis/tabs_v2_unified_data_analysis.py` (~1450 lines added)
   - `scripts/analysis/requirements-validation.txt` (girth, pingouin added)
   - The workflow file is unchanged vs main (the extras are unconditional, no CLI flag needed).

All numbers verified on frozen CRP-200; live N=297 produces same patterns or stronger.

### Multi-spec confirmatory factor analysis on Barriers
| Key | Description | Frozen N=200 result |
|---|---|---|
| `barriers_2f_cfa` | EFA-derived 2-factor (F1=14, F2=4) | CFI 0.871, RMSEA 0.072 (ML); CFI 0.967, RMSEA 0.073 (DWLS) |
| `barriers_3f_cfa` | Canonical F1a/F1b/F2 (best-fit) | CFI 0.886, RMSEA 0.069 (ML); **CFI 0.990, RMSEA 0.031 (DWLS)** |
| `barriers_4f_cfa` | Legacy 4-factor decomposition | CFI 0.827, RMSEA 0.085 |
| `barrier_model_comparison` | chi-squared difference + delta-AIC + delta-BIC across 1F/2F/3F/4F | 1F vs 2F p<.001; 2F vs 3F p<.001 |
| `joint_3construct_cfa` | Barriers + Readiness + Maturity in one CFA, free latent correlations | latent r(R,M) = 0.80 |
| `second_order_barriers_cfa` | F1a/F1b/F2 -> Barriers higher-order | DWLS CFI 0.990; F1a 0.87, F1b 1.00, F2 0.54 |

### Estimator-correct CFA for ordinal data
| Key | Description | Why |
|---|---|---|
| `cfa_dwls_estimator` | DWLS (WLSMV-equivalent) per construct | Mardia rejects multivariate normality; ML SEs biased |

### Bifactor analyses (general factor + specifics)
| Key | Description | Frozen N=200 result |
|---|---|---|
| `bifactor_rm` | R+M combined bifactor (25 items) | ECV 0.78, omega-h 0.85; specifics weak |
| `bifactor_barriers` | 18-item Barriers bifactor (G + F1aS/F1bS/F2S) | ECV 0.65, omega-h 0.79 |

### Item-level psychometrics
| Key | Description |
|---|---|
| `item_level_validity` | Hair convergent (>=0.50) + discriminant (cross < 0.30 OR gap >= 0.20) per item |
| `irt_grm` | Samejima graded response model: discrimination + thresholds + ceiling-effect flags |
| `dif_irt_smb_vs_ent` | Differential item functioning: per-item discrimination by SMB/Enterprise |
| `alpha_if_deleted_summary` | Items raising alpha if removed |
| `item_level_cohens_d_smb` | Per-item Cohen's d (SMB vs Enterprise) + Welch t |

### Subgroup analyses
| Key | Description |
|---|---|
| `subgroup_discriminant_validity` | HTMT bootstrap CI + Fornell-Larcker on F1a/F1b/F2 as 3 constructs |
| `subgroup_standalone_validation` | Per-subgroup standalone EFA + 1F CFA |
| `multigroup_3f_smb_vs_ent` | 3F Barriers CFA per group; latent correlations differ by org size |
| `reliability_by_demo` | Cronbach alpha per construct by SMB / Enterprise / For-Profit / Non-Profit |

### Inferential extensions
| Key | Description |
|---|---|
| `mediation_b_r_m` | Bootstrapped Barriers -> Readiness -> Maturity. Direct effect n.s.; indirect significant. |
| `per_factor_regressions` | Sub-factor regressions; reveals F2 sign-reversal |
| `standardized_subfactor_regressions` | Standardized betas + VIF (F1a/F1b/F2) |
| `equivalence_test_tost_smb_ent` | Two One-Sided Tests for equivalence (+/- 0.30 SD) |
| `power_analysis` | Smallest detectable Cohen's d at current N |

### Robustness + replication
| Key | Description |
|---|---|
| `mardia_normality` | Multivariate normality test (rejected -> use DWLS) |
| `mahalanobis_outliers` | Multivariate outlier counts at chi-squared p<.001 |
| `barriers_3f_cross_validation` | 50/50 split + Tucker congruence (frozen 0.984) |
| `bootstrap_alpha_ci` | 1000-iter bootstrap CIs on Cronbach's alpha |
| `esem_3factor` | ESEM with oblimin rotation (free cross-loadings); recommends B18 reassignment |

## Where each output surfaces on the website

The follow-up issue #1839 maps each new JSON key to existing pages on `/results/crp-2026/*` and `/results/*` plus how the React components share via the `variant` prop. **No new URLs.**

## Why unconditional (not gated)

Every daily run produces 28 keys. Both live and frozen CRP-200 datasets get the full statistical validation. Adds ~30 seconds to pipeline runtime; no compute concerns at this scale. Required for the website to surface complete validation evidence on every page refresh.

## Reference documents in workspace

- `03 Defense Presentation/Factor Analysis Defense Brief (5-1-2026 0834 EST).md`
- `03 Defense Presentation/Factor Analysis Extension Run (5-1-2026 0848 EST).md`
- `03 Defense Presentation/Per-Subgroup and Standalone Barriers Validation (5-1-2026 0956 EST).md`
- `03 Defense Presentation/Live N=297 Re-Run + Per-Factor Regressions (5-1-2026 1043 EST).md`
- `03 Defense Presentation/Tier 1+2 Findings - Mediation, Bifactor, DIF, ESEM (5-1-2026 1118 EST).md`
- `03 Defense Presentation/Future Analytic Improvements - Brainstorm (5-1-2026 1048 EST).md`


clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: `compare_cfa_models()` performs “nested chi-squared difference” tests for any pair where df decreases, but the 1F/2F/3F/4F barrier specifications aren’t guaranteed to be nested CFA models (different loading patterns). Reporting p-values here can be statistically invalid/misleading. At minimum, please clarify the nesting assumption in the output and/or restrict the chi-square diff to pairs you can justify as nested; otherwise rely on AIC/BIC deltas only.
2. `scripts/analysis/tabs_v2_validation.py`: Subgroup AVE lookup is using `safe_col`-renamed column names (`barrier_cols_list`) to index `barrier_loadings_matrix`, but EFA loadings are stored under the original (unsafe) column names from `run_efa()`. This will make all `lv` lookups return None, so `subgroup_aves` becomes all None and Fornell–Larcker results will be missing/incorrect. Consider normalizing the EFA loadings keys to safe names (or using original names for lookup via a safe->original map) before computing subgroup AVE.
```suggestion
    barrier_col_map = {c: safe_col(c) for c in BARRIER_COLS}
    barrier_cols_list = [barrier_col_map[c] for c in BARRIER_COLS]
    subgroup_aves = {}
    for grp_label, idxs in BARRIER_3GROUP.items():
        # F1a, F1b take F1 loadings (index 0); F2 takes F2 loadings (index 1)
        load_idx = 0 if grp_label in ('F1a', 'F1b') else 1
        lams = []
        for i in idxs:
            original_col = BARRIER_COLS[i]
            lv = barrier_loadings_matrix.get(original_col)
            if lv is not None and len(lv) > load_idx:
                lams.append(float(lv[load_idx]))
        subgroup_aves[grp_label] = ave_from_loadings(lams) if lams else None

    barrier_subgroup_data = df[BARRIER_COLS].rename(columns=barrier_col_map)
```
3. `scripts/analysis/tabs_v2_validation.py`: `all_safe_cols` is computed but never used in the joint 3-construct CFA block. This is dead code and can be removed to avoid confusion about which columns are actually used.
```suggestion

```
4. `scripts/analysis/tabs_v2_unified_data_analysis.py`: Subgroup AVE lookup has the same safe-name mismatch here: `barrier_loadings_matrix` keys come from EFA on the original columns, but `col` is taken from `barrier_cols_list = [safe_col(c) for c in BARRIER_COLS]`. That will make `lv` None for every item and will prevent `subgroup_aves` (and thus Fornell–Larcker) from being computed correctly. Consider key-normalizing the EFA loadings dict to safe names (or looking up by original column name).
5. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `barrier_1f_cfa` uses `setdefault('construct', 'Barriers_1F')`, but the existing 1F CFA dict already has `construct: 'Barriers'`, so it won’t be renamed. That makes the model-comparison output harder to interpret because the 1F model won’t be labeled consistently with `Barriers_2F/3F/4F`. Consider copying the dict and setting `construct` explicitly.
```suggestion
    # Copy it before relabeling so the original construct-level validation output
    # remains unchanged while the model-comparison view uses a consistent name.
    barrier_1f_cfa = dict(barrier_result.get('cfa', {}))
    barrier_1f_cfa['construct'] = 'Barriers_1F'
```
6. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `compare_cfa_models()` is described/structured as doing nested chi-square difference tests, but the 1F/2F/3F/4F barrier models are not guaranteed to be nested (they impose different loading patterns). Emitting p-values here can be statistically invalid/misleading. Consider restricting chi-square diff results to pairs you can justify as nested and otherwise using only ΔAIC/ΔBIC (or clearly labeling the chi-square diff as a heuristic, not a formal nested-model test).
7. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `compare_cfa_models()` uses `(m.get('chi2') or 0)` / `(m.get('df') or 0)` fallbacks when computing deltas. If any of these stats are None (e.g., missing from semopy output), this will silently produce incorrect chi-square differences, df differences, and AIC/BIC deltas. Consider skipping comparisons unless both models have non-null `chi2` and `df` (and only compute ΔAIC/ΔBIC when both are present).
```suggestion
            r_chi2 = r.get('chi2')
            full_chi2 = full.get('chi2')
            r_df = r.get('df')
            full_df = full.get('df')
            if r_chi2 is None or full_chi2 is None or r_df is None or full_df is None:
                continue
            d_chi2 = r_chi2 - full_chi2
            d_df = r_df - full_df
            if d_df <= 0:
                continue
            try:
                p = float(1 - chi2_dist.cdf(d_chi2, d_df))
            except Exception:
                p = None
            r_aic = r.get('aic')
            full_aic = full.get('aic')
            d_aic = round(r_aic - full_aic, 3) if r_aic is not None and full_aic is not None else None
            r_bic = r.get('bic')
            full_bic = full.get('bic')
            d_bic = round(r_bic - full_bic, 3) if r_bic is not None and full_bic is not None else None
            pairwise.append({
                'restricted': r.get('construct') or r.get('name'),
                'full': full.get('construct') or full.get('name'),
                'd_chi2': round(d_chi2, 3),
                'd_df': int(d_df),
                'p': round(p, 6) if p is not None else None,
                'd_aic': d_aic,
                'd_bic': d_bic,
```

clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: subgroup_standalone_validation() creates local_names but never uses it. This looks like leftover scaffolding and adds noise to the function; consider removing it or wiring it into the output if it’s meant to be exposed.
```suggestion

```
2. `scripts/analysis/tabs_v2_validation.py`: Per-subgroup standalone validation runs multiple heavy procedures (parallel_analysis with default n_iter=1000, plus EFA + CFA) across several subgroups. That’s a large runtime increase relative to the PR’s stated “6 extension analyses” scope and also grows the JSON output if persisted. If this isn’t required for #1836, consider gating it behind a flag or limiting it to --crp200 / a single sample.
3. `scripts/analysis/tabs_v2_unified_data_analysis.py`: compare_cfa_models() computes d_chi2 = restricted_chi2 - full_chi2 but does not guard against d_chi2 < 0 (which can happen if the “full” model fits worse or due to non-nested models). That yields misleading p-values (often ~1.0) and deltas. Consider skipping pairs where d_chi2 <= 0 (or setting p/d_chi2 to None and adding a note) so the output can’t be misinterpreted as a valid chi-square difference test.
4. `scripts/analysis/tabs_v2_unified_data_analysis.py`: subgroup_discriminant() computes subgroup means with DataFrame.mean(axis=1), which skips NaNs by default. For samples that include partial responses (e.g., v2_all), this can mix different item counts per row and make the reported Pearson correlations / Fornell–Larcker checks inconsistent with the listwise-complete EFA/AVE inputs. Consider using mean(axis=1, skipna=False) (or computing correlations from listwise-complete rows for each subgroup pair) to keep the correlation basis consistent.
```suggestion
    means = {label: data.mean(axis=1, skipna=False) for label, data in sub_data.items()}
```
5. `scripts/analysis/tabs_v2_unified_data_analysis.py`: run_validation() now computes subgroup_standalone_validation per sample (and does it for both canonical 3-group and legacy 4-group partitions). Each subgroup run includes parallel_analysis (n_iter=1000) and a CFA fit, so across 5 samples this can add substantial CPU time and risk slowing/timeout of the daily pipeline. Consider gating behind a CLI flag / limiting to primary_sample / lowering n_iter for subgroup runs.
```suggestion
    # These runs are expensive because each subgroup performs parallel analysis
    # and a one-factor CFA. Limit them to the primary sample to avoid multiplying
    # runtime across every sample in the daily pipeline.
    current_sample_key = locals().get('sample_key')
    primary_sample_key = locals().get('primary_sample')
    run_subgroup_standalone = (
        current_sample_key is None
        or primary_sample_key is None
        or current_sample_key == primary_sample_key
    )

    barrier_cols_safe = [safe_col(c) for c in BARRIER_COLS]
    barrier_renamed = df[BARRIER_COLS].rename(columns={c: safe_col(c) for c in BARRIER_COLS})
    legacy_4group = {label: idxs for label, idxs in BARRIER_SUBCONSTRUCTS.items()}

    if run_subgroup_standalone:
        print(f"\n{'='*70}")
        print(f"  PER-SUBGROUP STANDALONE VALIDATION (canonical 3-group F1a/F1b/F2)")
        print(f"{'='*70}")
        standalone_3group = subgroup_standalone_validation(
            barrier_renamed, BARRIER_3GROUP, barrier_cols_safe, BARRIER_NAMES
        )
        for r in standalone_3group:
            cfa = r.get('cfa_1f', {}) or {}
            v = r.get('verdict', {}) or {}
            print(f"  {r['name']:<20} k={r['k']} N={r['n_listwise']} alpha={r.get('alpha')} "
                  f"PA={r.get('parallel_analysis_factors')} CFI={cfa.get('cfi')} RMSEA={cfa.get('rmsea')} "
                  f"PASS={v.get('overall_pass')}")

        print(f"\n{'='*70}")
        print(f"  PER-SUBGROUP STANDALONE VALIDATION (legacy 4-group BARRIER_SUBCONSTRUCTS)")
        print(f"{'='*70}")
        standalone_4group = subgroup_standalone_validation(
            barrier_renamed, legacy_4group, barrier_cols_safe, BARRIER_NAMES
        )
        for r in standalone_4group:
            cfa = r.get('cfa_1f', {}) or {}
            v = r.get('verdict', {}) or {}
            print(f"  {r['name']:<35} k={r['k']} N={r['n_listwise']} alpha={r.get('alpha')} "
                  f"PA={r.get('parallel_analysis_factors')} CFI={cfa.get('cfi')} RMSEA={cfa.get('rmsea')} "
                  f"PASS={v.get('overall_pass')}")
    else:
        print(f"\n{'='*70}")
        print("  PER-SUBGROUP STANDALONE VALIDATION SKIPPED")
        print(f"{'='*70}")
        print(
            f"  Skipping expensive subgroup standalone validation for sample "
            f"{current_sample_key!r}; primary sample is {primary_sample_key!r}."
        )
        standalone_3group = []
        standalone_4group = []
```
6. `scripts/analysis/tabs_v2_unified_data_analysis.py`: These new extension-analysis helpers don’t appear to be covered by scripts/analysis/tests. Adding unit tests for at least key edge cases (e.g., negative d_chi2 handling in compare_cfa_models, NaN/partial-response handling in subgroup correlations, expected output shapes) would help prevent regressions in the daily pipeline outputs.
7. `scripts/analysis/tabs_v2_validation.py`: subgroup_discriminant() computes subgroup means with DataFrame.mean(axis=1), which skips NaNs by default. For samples that include partial responses (e.g., v2_all), this can mix different item counts per row and make the reported Pearson correlations / Fornell–Larcker checks inconsistent with the listwise-complete EFA/AVE inputs. Consider using mean(axis=1, skipna=False) (or computing correlations from listwise-complete rows for each subgroup pair) to keep the correlation basis consistent.
```suggestion
    means = {
        label: data.mean(axis=1, skipna=False)
        for label, data in sub_data.items()
    }
```

clarkemoyer:

1. `scripts/analysis/tabs_v2_unified_data_analysis.py`: For 1-factor EFA solutions, items have only one loading, so `secondary` becomes 0.0 and the current rules will always yield `discriminant='PASS'`. Consider setting `discriminant` (and `secondary`/`gap`) to `None` when `len(loads) < 2` so consumers don’t interpret DV as actually assessed.
2. `scripts/analysis/tabs_v2_validation.py`: `subgroup_discriminant()` computes subgroup means with `data.mean(axis=1)` (skipna default), which will include respondents with partial item missingness and makes the Pearson correlations non-listwise (and inconsistent with the unified pipeline version which uses `skipna=False`). Consider using `data.mean(axis=1, skipna=False)` (or explicitly dropping rows with any missing values for the subgroup) so subgroup correlations are computed on the same listwise-complete rows as the HTMT bootstrap and remain comparable across scripts.
```suggestion
    means = {label: data.mean(axis=1, skipna=False) for label, data in sub_data.items()}
```
3. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `subgroup_discriminant()` runs 3 HTMT bootstrap CIs at `n_boot=2000`. Because `run_validation()` is executed for multiple samples in the daily pipeline, this adds ~6000 additional bootstrap iterations per sample (doubling the existing HTMT bootstrap workload). Consider gating this subgroup HTMT/FL block behind `primary_sample` (like the standalone subgroup validation) and/or reducing `n_boot` for non-primary samples to keep daily runs from slowing down significantly.
4. `scripts/analysis/tabs_v2_validation.py`: When EFA retains a single factor, each item only has one loading, so `secondary` is forced to 0.0 and the current logic will always mark `discriminant='PASS'`. Consider emitting `discriminant=None` (and `secondary`/`gap` as `None`) when `len(loads) < 2` so the output doesn’t imply DV was evaluated for 1-factor solutions.

clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: This block adds a large set of additional analyses (DWLS CFAs, bifactor/second-order models, Mardia/Mahalanobis, cross-validation, IRT GRM, regressions) and later emits them as new top-level JSON keys. That scope appears to go beyond the PR/issue description of “6 defense-ready analyses” (plus alpha-if-deleted). Consider gating these extras behind an explicit CLI flag (or moving them to a follow-up PR) to keep the production validation pipeline’s scope and runtime aligned with the stated goal.
2. `scripts/analysis/requirements-validation.txt`: requirements-validation.txt is modified in this PR, but the PR description states the diff against main is “exactly the two real script files”. Please either update the PR description to include this dependency change or revert it if it’s not intended to ship with this PR.
3. `scripts/analysis/tabs_v2_unified_data_analysis.py`: run_validation() now runs multiple semopy DWLS fits, bifactor/second-order models, Mardia/Mahalanobis, IRT, and regressions for every sample in SAMPLE_ORDER. In the daily pipeline this can multiply runtime significantly (up to 5 samples). Consider running these “Linderman-grade”/IRT/regression extensions only for the primary sample (similar to subgroup_standalone_validation), or adding a CLI flag to disable them in the daily run.
```suggestion
    # IRT graded response models and per-factor regressions are "Linderman-grade"
    # extensions that are materially more expensive than the core validation steps.
    # Run them only for the primary sample so daily multi-sample validation does not
    # multiply runtime across every entry in SAMPLE_ORDER.
    irt_results = {}
    per_factor_reg = {}
    if primary_sample:
        barrier_short_ids = [f'B{i+1}' for i in range(len(BARRIER_NAMES))]
        readiness_short_ids = [f'R{i+1}' for i in range(len(READINESS_NAMES))]
        maturity_short_ids = [f'M{i+1}' for i in range(len(MATURITY_NAMES))]
        irt_results['Barriers'] = irt_grm(barrier_renamed, barrier_cols_safe, barrier_short_ids, BARRIER_NAMES)
        irt_results['Readiness'] = irt_grm(readiness_renamed, [safe_col(c) for c in READINESS_COLS], readiness_short_ids, READINESS_NAMES)
        irt_results['Maturity'] = irt_grm(maturity_renamed, [safe_col(c) for c in MATURITY_COLS], maturity_short_ids, MATURITY_NAMES)
        if 'error' not in irt_results['Barriers']:
            print(f"  IRT Barriers: mean discrimination={irt_results['Barriers'].get('mean_discrimination')}, "
                  f"ceiling-effect items={irt_results['Barriers'].get('ceiling_effect_count')}")

        # Per-factor regressions (sub-factor decomposition vs full-scale aggregate)
        per_factor_reg = per_factor_regressions(df, BARRIER_COLS, READINESS_COLS, MATURITY_COLS, BARRIER_3GROUP)
        if 'Readiness_mean' in per_factor_reg and 'error' not in per_factor_reg.get('Readiness_mean', {}):
            rr = per_factor_reg['Readiness_mean']
            print(f"  Per-factor Reg (Readiness): R^2 total={rr['total_scale_model']['r2']} "
                  f"-> sub-factor={rr['sub_factor_model']['r2']} (lift={rr['r2_lift_from_decomposition']})")
    else:
        irt_results = {
            'skipped': True,
            'reason': 'extended_validation_only_runs_for_primary_sample'
        }
        per_factor_reg = {
            'skipped': True,
            'reason': 'extended_validation_only_runs_for_primary_sample'
        }
```
4. `scripts/analysis/tabs_v2_unified_data_analysis.py`: subgroup_standalone_validation() calls parallel_analysis() without overriding n_iter, so it uses the default 1000 iterations. That’s substantially more expensive than the standalone validator’s subgroup check (which uses fewer iterations) and will slow the daily pipeline when primary_sample=True. Consider passing a smaller n_iter (e.g., 200) here as well.
5. `scripts/analysis/tabs_v2_validation.py`: The module docstring’s analysis list/usage block is now out of sync with what the script actually runs: the file implements additional sections beyond #16 (e.g. subgroup standalone validation, DWLS/bifactor/IRT/regressions) and main() supports a --crp200 flag but the top-level Usage line doesn’t mention it. Please update the docstring so it accurately reflects the supported CLI flags and the analyses that will appear in the JSON output.
6. `scripts/analysis/tabs_v2_validation.py`: There’s a duplicated/garbled inline comment here (two headings on the same line). Please clean this up so the section header is readable in logs and doesn’t look like an accidental paste.
```suggestion
    # -- Per-subgroup standalone validation (does each barrier subgroup hold as its own scale?) --
```

clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: `means[n1][mask].corr(means[n2][mask])` can return NaN (e.g., <2 valid rows after masking or constant series). Because `json.dump()` serializes float NaN as `NaN` (invalid JSON for `JSON.parse`/schema validation), this should be normalized to `None` before rounding/storing (and `abs(r)` / Fornell–Larcker should handle the missing correlation as well).
2. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `means[n1][mask].corr(means[n2][mask])` can produce NaN (too few rows after masking or zero-variance series). Since Python `json.dump()` will emit float NaN as `NaN` (invalid JSON for downstream `JSON.parse`/Zod validation), normalize `r` to `None` when it’s NaN and avoid using `abs(r)` in the Fornell–Larcker block when `r` is missing.
```suggestion
            raw_r = means[n1][mask].corr(means[n2][mask])
            r = None if pd.isna(raw_r) else float(raw_r)
            pearson_corr[f'{n1} vs {n2}'] = round(r, 4) if r is not None else None
            ave1 = group_aves.get(n1)
            ave2 = group_aves.get(n2)
            if ave1 is not None and ave2 is not None:
                sa1 = math.sqrt(ave1)
                sa2 = math.sqrt(ave2)
                small = min(sa1, sa2)
                abs_r = abs(r) if r is not None else None
                fl_results.append({
                    'pair': f'{n1} vs {n2}',
                    'sqrt_ave_1': round(sa1, 4),
                    'sqrt_ave_2': round(sa2, 4),
                    'abs_r': round(abs_r, 4) if abs_r is not None else None,
                    'smaller_sqrt_ave': round(small, 4),
                    'pass': bool(small > abs_r) if abs_r is not None else None,
```
3. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `subgroup_standalone_validation()` calls `parallel_analysis()` without overriding `n_iter`, so it uses the default 1000 iterations per subgroup. Since this block is explicitly described as “computationally expensive”, consider passing a lower `n_iter` (e.g., 200) for these subgroup checks to keep primary-sample runtime bounded while still providing a stable diagnostic.
```suggestion
            n_factors_pa = parallel_analysis(data, max_factors=min(4, k), n_iter=200)
```
4. `scripts/analysis/tabs_v2_unified_data_analysis.py`: The “LINDERMAN-GRADE EXTENSIONS” block runs automatically for `primary_sample` with no CLI flag, and its results are also emitted into the validation JSON (`cfa_dwls_estimator`, `bifactor_rm`, `second_order_barriers_cfa`, etc.). This looks broader than the PR description’s 6 defense-ready extension analyses; if these extras aren’t intended for the production daily pipeline, gate them behind an explicit CLI flag (similar to `tabs_v2_validation.py --linderman`) to avoid unintentional runtime increases / output-surface changes.

clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: The JSON output block unconditionally references `mediation_results`, `std_reg`, `bootstrap_alpha_results`, `item_d_smb`, etc., but those variables are only defined inside the `if use_linderman:` block above. Running with `--json` *without* `--linderman` will raise a `NameError`. Initialize these variables to a default (e.g., skipped/error dicts) before the `if use_linderman:` block, or only emit these keys when `use_linderman` is enabled.
```suggestion
            output['mediation_b_r_m'] = mediation_results
            output['standardized_subfactor_regressions'] = std_reg
            output['bootstrap_alpha_ci'] = bootstrap_alpha_results
            output['item_level_cohens_d_smb'] = item_d_smb
            output['reliability_by_demo'] = reliability_demo
            output['power_analysis'] = power_results
            output['equivalence_test_tost_smb_ent'] = tost_results
            output['bifactor_barriers'] = bifactor_b_results
            output['multigroup_3f_smb_vs_ent'] = multigroup_3f_results
            output['dif_irt_smb_vs_ent'] = dif_results
            output['esem_3factor'] = esem_results
            output['discriminant_validity'] = discrim
```
2. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `barrier_renamed` and `barrier_cols_safe` are used in the Tier 1+2 extensions block (`bifactor_barriers`, `multigroup_3f_sem`, `esem_target_rotation`), but they are only assigned later inside `if primary_sample:`. For non-primary samples (the common case in the daily pipeline), this will raise `NameError` and abort validation. Define `barrier_cols_safe` / `barrier_renamed` immediately after `safe_barrier_data` is created (or reuse `safe_barrier_data` + `barrier_cols_list`) so they exist for both primary and non-primary samples.
3. `scripts/analysis/requirements-validation.txt`: PR description says the diff against main is exactly two script files, but this PR also modifies `scripts/analysis/requirements-validation.txt` (adds new dependencies). Please update the PR description to match, or drop this file change if it’s unintended.

clarkemoyer:

clarkemoyer:

1. `scripts/analysis/tabs_v2_unified_data_analysis.py`: `run_validation(..., primary_sample=...)` gates several of the new outputs behind `if primary_sample:` (e.g., `cfa_dwls_estimator`, `bifactor_rm`, `mardia_normality`, `irt_grm`, `per_factor_regressions`). This means non-primary sample tiers won't have the full set of new keys, which conflicts with the PR/issue description that all 28 new analyses run for every sample tier on each daily invocation, and may break the sample dropdown if the UI expects these keys on all samples. Consider either (a) computing these blocks for every sample tier, or (b) emitting the keys for all samples with an explicit `{skipped: true, reason: ...}` sentinel so the schema stays consistent.
2. `scripts/analysis/tabs_v2_unified_data_analysis.py`: The `primary_sample` docstring says non-primary samples only skip the per-subgroup standalone validation, but the implementation also reduces bootstrap iterations (`n_boot=...`) and skips the entire extended psychometric block (DWLS CFA, bifactor, normality/outliers, cross-validation, IRT, per-factor regressions). Update the docstring (and/or parameter name) so callers understand which outputs are omitted when `primary_sample=False`.
3. `scripts/analysis/verify_against_R.R`: `spearman_brown_split_half` is computed as `2 * average_r / (1 + average_r)`, which uses the average inter-item correlation from `psych::alpha` rather than an actual split-half correlation between two test halves. If this field is meant to cross-check Python’s `split_half_reliability()` (odd-even split + Spearman–Brown), compute `r` as the correlation of two halves (e.g., odd vs even totals, or `psych::splitHalf`) and then apply the prophecy formula; otherwise rename this field to avoid implying it is a split-half estimate.

clarkemoyer:

1. `scripts/analysis/tabs_v2_validation.py`: The header comment says the advanced extensions “always run on every invocation”, but later in main() several of those blocks are explicitly gated (e.g., per-subgroup standalone validation runs only when --crp200 is set). Please either adjust the header wording to reflect the actual gating, or remove the gating so runtime behavior matches the documentation.

2. `scripts/analysis/tabs_v2_validation.py`: subgroup_aves for the F1a/F1b/F2 Fornell–Larcker check is currently derived from the *2-factor EFA* loadings (F1 used for both F1a and F1b, F2 for F2). That doesn’t align with the stated intent of treating F1a/F1b/F2 as three constructs, and it also diverges from the available 3F CFA loadings you compute just above. Consider computing AVE per subgroup from the barriers_3f_cfa standardized loadings when semopy results are available (and only fall back to the 2F EFA loadings if CFA is unavailable).

3. `scripts/analysis/tabs_v2_unified_data_analysis.py`: run_validation() now skips many of the new “extended” analyses for non-primary samples (emitting a {"skipped": true} sentinel). The PR description/issue text indicates the 28 new keys are produced unconditionally for every daily run and (per #1836/#1839) expected to exist at every sample tier in the unified validation output. Please either (a) remove/relax the primary_sample gating so non-primary samples get real results too, or (b) update the PR description and ensure downstream consumers/UI explicitly handle the sentinel values when users switch sample tiers.
4. `scripts/analysis/tabs_v2_unified_data_analysis.py`: subgroup_aves for the subgroup Fornell–Larcker block is derived from the 2-factor EFA loadings (factor 0 reused for both F1a and F1b). Since this block is explicitly treating F1a/F1b/F2 as three constructs, AVE should ideally come from the 3F CFA standardized loadings (barriers_3f_cfa) when available, with EFA as a fallback. Otherwise the Fornell–Larcker comparison can be misleading/mismatched to the stated construct definitions.


clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The BOOTSTRAP command includes `INPUT=` with no variable list. As written, the next token is `/CRITERIA`, which is likely to be parsed as part of `INPUT=` and can make the syntax fail to run. If no input variables are needed, drop the `INPUT=` clause entirely (or provide an explicit variable list per SPSS BOOTSTRAP syntax).
2. `scripts/spss/tabs_v2_validation.sps`: Same issue here: `INPUT=` is present without any variables, so the following `/CRITERIA` line will likely be parsed incorrectly and break execution. Remove `INPUT=` or provide the intended input variable list.
3. `scripts/spss/tabs_v2_validation.sps`: Same `BOOTSTRAP ... INPUT=` issue: there is no variable list after `INPUT=`, which can cause a syntax error when `/CRITERIA` follows immediately. Remove `INPUT=` (or fill it in) so the syntax runs end-to-end.
4. `scripts/spss/tabs_v2_validation.sps`: In the Barriers EFA FACTOR block, `/CRITERIA` is specified twice. Depending on SPSS's parsing rules, the second `/CRITERIA ITERATE(50)` can override the earlier `/CRITERIA FACTORS(2) ...` and unintentionally drop the fixed factor count. Remove the duplicate `/CRITERIA` line (or merge options into a single `/CRITERIA` clause).


clarkemoyer:

1. `scripts/analysis/tests/test_parity_to_published_formulas.py`: The docstring’s suggested command (`python -m pytest test_parity_to_published_formulas.py -v`) only works if you run from the tests directory. Consider updating it to a repo-root-safe invocation (e.g., `pytest scripts/analysis/tests/test_parity_to_published_formulas.py -v`) to prevent confusion.

2. `scripts/cross-platform-verification/build_zip.py`: `SCRIPTS` is defined but never used. Please remove it (or use it) to avoid dead code and keep the script lint-clean.


clarkemoyer:

1. `scripts/minitab/tabs_v2_validation.MTB`: The header comment claims SMB_ENT is n~46 vs n~131, which conflicts with the SPSS syntax comment (n~114 vs n~86) and the PR description’s group sizes for CRP-200. Please correct the expected n’s here (or clarify the exact filtering that yields these counts) so users can use this as a reliable spot-check.

2. `scripts/minitab/build_minitab_csv.py`: `org_grp()` returns empty strings for unknown/missing org sizes, which makes `SMB_ENT` an object/mixed-type column in pandas. That can lead some CSV importers to treat the column as text rather than numeric (and Minitab users then won’t be able to use it as a grouping variable without manual conversion). Prefer returning `np.nan` for missing/unknown and keeping the column numeric; `to_csv(..., na_rep='')` will still render blanks for Minitab.
3. `scripts/spss/build_spss_artifacts.py`: The value labels written into the .sav are shortened (e.g., Barriers value 2 is labeled "Minor" instead of "Minor Barrier"; Readiness value 5 is "Very High Readiness" vs the source label "Very High Readiness/Capability"). Since the docstring says the .sav embeds Likert anchors, it’s better to keep these labels identical to the source dataset/pipeline mapping to avoid confusion when interpreting SPSS output.

4. `scripts/spss/tabs_v2_validation.sps`: The targeted license list includes "Statistics Base" twice. Consider deduplicating the list to avoid confusion about which modules are actually required for the syntax in this folder.


clarkemoyer:

1. `scripts/spss/README.md`: The blockquote in the recommended dissertation paragraph has a broken line break (`>` on its own line) that turns “Bootstrapping + Missing Values modules” into a separate list item inside the quote. This renders oddly in Markdown and makes the suggested paragraph harder to copy/paste. Consider rewriting this sentence so the module list is on a single quoted line (or remove the blank `>` line).

2. `scripts/spss/build_spss_artifacts.py`: `pyreadstat` is imported at module import time, so running this script without the dependency installed fails with a raw ImportError before any helpful guidance is printed. Since this is a one-off regeneration tool, consider importing `pyreadstat` inside `main()` with a try/except and exiting with a clear message (e.g., how to install the dependency).


clarkemoyer:

clarkemoyer:

1. `scripts/cross-platform-verification/build_zip.py`: In the Python bundle README text, the file list says `README_PYTHON.md <- this file`, but `build_one()` always writes the top-level readme as `README.md`. This is confusing for consumers of the zip; either update the listing to `README.md` or write the Python readme under the name `README_PYTHON.md` and keep the list consistent.

2. `scripts/analysis/verify_against_R.R`: `encode_scale()` silently leaves values as NA when the label isn't in `scale_map` (there’s no `else` branch). Since the header claims this mirrors the Python encoder, it would be safer to fail fast on unknown labels (e.g., `stop()` with the unexpected value) so the verification run can’t accidentally succeed on mis-encoded data.

3. `scripts/analysis/verify_against_R.R`: The final message says all values should agree to 4 decimal places, but earlier in this repo’s methodology notes you call out expected small differences (e.g., listwise vs pairwise scope) and tolerances like <=0.014. Consider adjusting this output text to match the actual acceptance criteria so users don’t treat small expected deltas as failures.

4. `scripts/minitab/tabs_v2_validation.MTB`: The validation pipeline’s EFA uses Promax rotation for multi-factor solutions (see `scripts/analysis/tabs_v2_validation.py`), but this Minitab macro runs Barriers EFA with `Rotation Varimax`. That will change the loading matrix and can create false discrepancies when cross-checking. If Minitab supports Promax in session commands, switch to it (and match the kappa/exponent); otherwise, clarify in the header that exact loading parity requires running the GUI dialog with Promax enabled.
5. `scripts/minitab/README.md`: This table claims the Barriers EFA command with `Rotation Varimax` “matches Python loadings”, but the Python pipeline uses Promax for multi-factor EFA. To avoid misleading users, either change the recommended rotation here to Promax (kappa=4) or adjust the wording to note that Varimax won’t reproduce the pipeline’s oblique loadings.

6. `scripts/cross-platform-verification/build_zip.py`: For the Python bundle, `requirements.txt` is populated from `scripts/analysis/requirements-validation.txt` only, but that file explicitly depends on the base `scripts/analysis/requirements.txt` (numpy/pandas/scipy/etc.). As-is, `pip install -r requirements.txt` in the unzipped bundle won’t install the core deps that the README claims are included. Consider either bundling a combined requirements file (base + validation) or include both files and update the quick-start accordingly.


clarkemoyer:
**Copilot Review Cycle (Round 1/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The SPSS BOOTSTRAP blocks use NSAMPLES=2000, but the canonical Python `bootstrap_alpha_ci` defaults to 1000 resamples. If the intent is for SPSS bootstrap CIs to match Python “1:1”, align the resample count (or update the comments/docs to avoid claiming exact parity when the sampling scheme differs).
2. `scripts/spss/README.md`: This README states the bootstrap percentile CIs around Cronbach’s alpha “match Python `bootstrap_alpha_ci`”, but the bundled SPSS syntax uses 2000 bootstrap samples while Python defaults to 1000. Either align the resample count across tools or clarify that results should be close but not expected to match exactly.

3. `scripts/cross-platform-verification/build_zip.py`: PR description references a single shareable zip named `tabs_v2_cross_platform_verification.zip`, but this script (and the committed artifacts) produce three per-tool bundles named `tabs_v2_validation_{spss|minitab|python}.zip`. Please align the PR description (or rename the artifacts) so reviewers/users aren’t looking for a file that doesn’t exist.

clarkemoyer:

PR #1838, opened 2026-05-01
feat(pipeline): 21-day auto-approve runway tracking + fix 3 daily-report regressions

clarkemoyer:
Two related changes ship together because the daily report needs the new runway data to render the HIGH RISK section, and the report had three pre-existing render bugs that were already affecting every daily issue.

## Pipeline data additions

Prolific auto-approves any AWAITING REVIEW submission **21 days** after \`completed_at\`, regardless of TABS quality flags. Until now the pipeline discarded that timestamp, so no script could see how much runway each submission had. Threading it through the pipeline lets every consumer (messaging, rejection, triage, reports) reason about deadlines without making its own API call.

| # | File | Change |
|---|---|---|
| 1 | \`scripts/analysis/tabs_api.py\` | New \`prolific_submission_summaries()\` returns \`{PID: {status, completed_at, started_at}}\`. Existing \`prolific_submission_statuses()\` kept for backward compat. |
| 2 | \`scripts/analysis/fetch_prolific_data.py\` | Writes the richer dict to \`statuses.json\`. |
| 3 | \`scripts/analysis/enrich_qualtrics_csv.py\` | Adds \`Prolific_Completed_At\` + \`Prolific_Started_At\` columns. Loader normalises legacy + current shapes. |
| 4 | \`scripts/analysis/tabs_v2_data_audit.py\` + \`disposition_triage.py\` | New columns flow through to \`disposition.csv\`. |
| 5 | \`scripts/analysis/generate_disposition_summary.py\` | New \`autoApproveRunway\` block in \`disposition-summary.json\` with 4 buckets (Comfortable / Monitor / Act soon / **HIGH RISK**) + top-level \`highRiskCount\` and \`highRiskByDisposition\` for downstream consumers. |

**Privacy:** only aggregate counts land in the committed JSON. Per-PID \`Prolific_Completed_At\` lives only in the \`disposition-csv\` artifact (1-day retention).

## Daily report regressions fixed

Audit of issues #1823 / #1826 / #1829 / #1832 / #1835 showed three sections rendering wrong on **every daily report for >5 days**:

| Fix | Section | Root cause |
|---|---|---|
| **A** | Disposition Triage table empty | Reads \`triage-metrics/*.txt\` artifact that no pipeline step writes. Added fallback computing total + breakdown from \`dashboard-data/disposition-summary.json\`. **(Same fix as #1700 — supersedes that PR.)** |
[...]

PR #1842, opened 2026-05-01
fix(stale-triage): fall back to ObjectId timestamp when sent_at is null

clarkemoyer:
## Bug

Prolific's messages API returns \`sent_at\` and \`created_at\` as null/undefined for participant↔researcher messages in this study. \`find_stale_return_requested._msg_time()\` returned None for those messages, causing two failures:

1. **35 of 40 AWAITING REVIEW PIDs were invisible to today's stale-triage** — they had researcher messages on file but with null timestamps, so the script returned \`(None, None)\` and never bucketed them.
2. **Participant replies were silently dropped.** A PID with \`return_requested\` set + a participant reply (with null \`sent_at\`) was misclassified as \`stale_no_reply_to_rr\` and queued for rejection, even though the participant had responded.

Discovered while auditing the 8 rejections from 2026-05-01 morning. The audit (8 dispatches of \`single\` mode in \`check-participant.yml\`) confirmed:
- All 8 of the rejected PIDs were validly rejected (no participant messages on any of them — 3 messages each, all from researcher \`68264cbfdeb62546fe6060fe\`).
- But PID \`69dac332cb565c58081269a6\` (still AWAITING REVIEW, FLAG-SINGLE-IRI) **DID reply** — *\"Yes, that must have been an oversight. Because I take my time and read each question carefully. Thank you.\"* — and is currently invisible to the stale-triage because both messages have \`sent_at: undefined\`.

## Fix

Extend \`_msg_time()\` with a third fallback that decodes the embedded Unix epoch from the message id's MongoDB ObjectId prefix (first 4 bytes = epoch seconds). MongoDB ObjectIds always carry a creation timestamp by construction, so this is the same anchor the Prolific server uses internally — authoritative, deterministic, no extra API call.

Resolution order:
1. \`sent_at\` (explicit server-recorded send time)
2. \`created_at\` (explicit server-recorded creation time)
3. **NEW**: ObjectId timestamp prefix from \`id\`

## Tests

10 new tests in \`TestObjectIdFallback\`:

- \`_objectid_time\` decoder: real wild-collected id (\`69ef4762...\` → 2026-04-27 11:24:18 UTC), non-string input, short string, non-hex prefix.
- \`_msg_time\` resolution order: explicit \`sent_at\` wins over ObjectId; ObjectId used when explicit timestamps absent.
- \`classify_submission\` now correctly excludes a PID from \`stale_no_reply_to_rr\` when the participant's reply id encodes a timestamp after the \`return_requested\` anchor — the exact bug pattern from this morning's investigation.

Existing tests still pass: the message in \`test_classify_tolerates_timestampless_participant_message\` has bad sent_at AND no \`id\`, so the ObjectId fallback also returns None — same outcome as before.

## Test plan

- [x] \`pytest scripts/analysis/tests/test_find_stale_return_requested.py\` — 29/29 pass (was 19, +10 new)
- [x] Full Python suite: 429 passed (vs 419 baseline), no new failures (52 pre-existing Windows-only encoding issues that don't fire in CI)
- [x] \`npm run format\` / \`lint\` / \`test\` (582/582) / \`build\` — all clean
- [x] Audit confirmed today's 8 rejections were valid (no false rejections to undo)

## What this PR does NOT do

- Does **not** unreject anyone — audit confirmed the 8 rejections this morning were correct.
- Does **not** change the daily-report layout — the bucket numbers will populate accurately starting tomorrow's run.

## Related

- Companion to #1838 (which surfaced this issue while building the 21-day auto-approve runway tracker)

### DIFF
--- a/scripts/analysis/find_stale_return_requested.py
+++ b/scripts/analysis/find_stale_return_requested.py
@@ -116,7 +116,12 @@ def _objectid_time(message_id):
     if not isinstance(message_id, str) or len(message_id) != 24:
         return None
     try:
-        int(message_id, 16)  # validate all 24 chars are hex
+        raw = bytes.fromhex(message_id)  # strict hex-only, rejects underscores
+    except ValueError:
+        return None
+    if len(raw) != 12:
+        return None
+    try:
         ts = int(message_id[:8], 16)
     except ValueError:
         return None
--- a/scripts/analysis/find_stale_return_requested.py
+++ b/scripts/analysis/find_stale_return_requested.py
@@ -97,17 +97,54 @@ def _parse_iso(ts):
         return None
 
 
-def _msg_time(m):
-    """Return the best-effort timestamp for a message, or ``None`` if neither
-    ``sent_at`` nor ``created_at`` is a parseable timestamp.
+def _objectid_time(message_id):
+    """Decode the embedded Unix timestamp from a 24-char hex MongoDB ObjectId.
 
-    Prolific returns a small number of messages with both fields null or
-    malformed (observed in live data: ~4 per daily run out of ~560 subs).
-    These are typically system-generated records. Returning ``None`` lets
-    callers skip the message for ordering purposes rather than failing the
-    entire PID's classification.
+    Prolific message IDs are MongoDB ObjectIds whose first 4 bytes (8 hex
+    chars) encode the creation time as a Unix epoch. When ``sent_at`` and
+    ``created_at`` are both absent or unparseable from the messages API
+    response (observed across the bulk of FLAG-* / clarification messages,
+    not just a few system records), this gives us a deterministic fallback
+    ordering anchor — the same anchor the Prolific server itself uses
+    internally for message creation.
+
+    Reference: https://www.mongodb.com/docs/manual/reference/method/ObjectId/
+
+    Returns ``None`` for inputs that don't look like a 24-char hex string,
+    so callers can chain it after the explicit-timestamp fields.
+    """
+    if not isinstance(message_id, str) or len(message_id) < 8:
+        return None
+    try:
+        ts = int(message_id[:8], 16)
+    except ValueError:
+        return None
+    if ts <= 0:
+        return None
+    try:
+        return datetime.fromtimestamp(ts, tz=timezone.utc)
+    except (OSError, OverflowError, ValueError):
+        return None
+
+
+def _msg_time(m):
+    """Return the best-effort timestamp for a message, or ``None`` if no
+    fallback yields a parseable instant.
+
+    Resolution order (most authoritative first):
+      1. ``sent_at`` — explicit server-recorded send time
+      2. ``created_at`` — explicit server-recorded creation time
+      3. The 4-byte timestamp prefix of the message ``id`` (MongoDB
+         ObjectId convention) — used because the Prolific messages API
+         frequently returns ``sent_at`` and ``created_at`` as null/undefined
+         for participant↔researcher messages, which previously caused this
+         function to drop the message entirely. Dropping a participant
+         message is consequential: the caller treats the PID as
+         "no reply since RR" and queues it for rejection.
     """
     t = _parse_iso(m.get("sent_at")) or _parse_iso(m.get("created_at"))
+    if t is None:
+        t = _objectid_time(m.get("id"))
     return t
 
 
--- a/scripts/analysis/find_stale_return_requested.py
+++ b/scripts/analysis/find_stale_return_requested.py
@@ -125,7 +125,7 @@ def _objectid_time(message_id):
         ts = int(message_id[:8], 16)
     except ValueError:
         return None
-    if ts <= 0:
+    if ts < 0:
         return None
     try:
         return datetime.fromtimestamp(ts, tz=timezone.utc)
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `scripts/analysis/find_stale_return_requested.py`: _objectid_time() docstring says it returns None for inputs that don't look like a 24-char hex ObjectId, but the implementation only requires len>=8 and only validates the first 8 chars. This can misclassify non-ObjectId IDs that happen to start with 8 hex chars (e.g., an API format change like '69ef4762-...'), potentially excluding a truly stale PID from triage. Consider requiring exactly 24 hex chars (and ideally validating the full string) before decoding the prefix; add a test for a >=8 but non-24 id to lock in the behavior.

clarkemoyer:

1. `scripts/analysis/find_stale_return_requested.py`: `int(message_id, 16)` is used to “validate all 24 chars are hex”, but Python’s `int()` accepts underscores in numeric strings (e.g. `"ff_ff"`), so this isn’t a strict hex-bytes validation as the comment/docstring implies. Consider validating with `bytes.fromhex(message_id)` (and ensuring it yields 12 bytes) or a strict `[0-9a-fA-F]{24}` check before decoding the timestamp prefix.
```suggestion
        raw = bytes.fromhex(message_id)
        if len(raw) != 12:
            return None
```
2. `scripts/analysis/tests/test_find_stale_return_requested.py`: `test_objectid_decoder_returns_none_for_non_hex_prefix` currently passes a 16-character string, so it returns `None` due to the length guard rather than exercising the non-hex validation path. Use a 24-character string with a non-hex character (or rename the test) so the test actually covers the intended behavior.
```suggestion
        assert _objectid_time("zzzzzzzzdeadbeefdeadbeef") is None
```
3. `scripts/analysis/tests/test_find_stale_return_requested.py`: This test no longer asserts anything about the returned `record`. If `classify_submission` regresses and starts returning `(bucket, None)` here, the test would still pass. Consider restoring a minimal assertion (e.g., `record is not None` and `record["pid"] == "PID_T"`) to keep the test consistent with the rest of the suite.
```suggestion
        assert bucket == "stale_no_reply_to_message"
        assert record is not None
        assert record["pid"] == "PID_T"
```

clarkemoyer:

1. `scripts/analysis/tests/test_find_stale_return_requested.py`: The test embeds a real-looking 24-char hex identifier ("69ef4762f0591a0ae79ea82f") and calls it a "real Prolific message id observed in the wild". In this repo, 24-char hex strings are also the shape of Prolific participant IDs (PII), so committing any real captured identifiers is risky. Please replace this with a clearly synthetic ObjectId constructed from a known epoch (as you already do in the later test) and update the docstring to avoid implying it came from production data.
2. `scripts/analysis/tests/test_find_stale_return_requested.py`: This test repeats the same real-looking 24-char hex id literal. To reduce any chance of leaking real identifiers (and to make the test intent clearer), define a synthetic ObjectId test fixture once (derived from a chosen epoch + fixed suffix) and reuse it here instead of hardcoding a value that could be mistaken for a Prolific PID.

clarkemoyer:

1. `scripts/analysis/find_stale_return_requested.py`: `_objectid_time()` rejects ObjectIds whose embedded epoch seconds are `0` via `if ts <= 0: return None`. A timestamp of 0 (1970-01-01T00:00:00Z) is valid per the ObjectId spec and would currently decode incorrectly. Consider allowing `0` (e.g., `ts < 0` or removing this guard) and rely on the existing length/hex/fromtimestamp checks for validation.

clarkemoyer:

|---|---|---|---|---|
| 1 | 1 | find_stale_return_requested.py | 🔧 Fixed | 2026-05-01 19:57 |
| 2 | 3 | find_stale_return_requested.py, test_find_stale_return_requested.py | 🔧 Fixed | 2026-05-01 20:09 |
| 3 | 2 | test_find_stale_return_requested.py | 🔧 Fixed | 2026-05-01 20:22 |
| 4 | 1 | find_stale_return_requested.py | 🔧 Fixed | 2026-05-01 20:34 |
| 5 | 0 | — | ✅ Clean | 2026-05-01 20:45 |

### Round 1 Comments (1)

- `scripts/analysis/find_stale_return_requested.py` — _objectid_time() docstring says it returns None for inputs that don't look like a 24-char hex ObjectId, but the implemen

### Round 2 Comments (3)

- `scripts/analysis/find_stale_return_requested.py` — `int(message_id, 16)` is used to “validate all 24 chars are hex”, but Python’s `int()` accepts underscores in numeric st
- `scripts/analysis/tests/test_find_stale_return_requested.py` — `test_objectid_decoder_returns_none_for_non_hex_prefix` currently passes a 16-character string, so it returns `None` due
- `scripts/analysis/tests/test_find_stale_return_requested.py` — This test no longer asserts anything about the returned `record`. If `classify_submission` regresses and starts returnin

### Round 3 Comments (2)

- `scripts/analysis/tests/test_find_stale_return_requested.py` — The test embeds a real-looking 24-char hex identifier ("69ef4762f0591a0ae79ea82f") and calls it a "real Prolific message
- `scripts/analysis/tests/test_find_stale_return_requested.py` — This test repeats the same real-looking 24-char hex id literal. To reduce any chance of leaking real identifiers (and to

### Round 4 Comments (1)

- `scripts/analysis/find_stale_return_requested.py` — `_objectid_time()` rejects ObjectIds whose embedded epoch seconds are `0` via `if ts <= 0: return None`. A timestamp of 

### Result

✅ All automated review comments addressed in 5 round(s). Ready for final human review.


**Copilot Review Cycle:** ✅ Passed after 5 round(s) - review clean, all CI checks green. Ready to merge.

PR #1860, opened 2026-05-02
feat(spss): one-click double-click launcher for SPSS validation bundle

clarkemoyer:
## Summary

Make the SPSS cross-platform verification bundle a true double-click experience for committee members and other recipients receiving the zip via email. End-to-end tested by the author against a real SPSS Statistics 31.0 install on Windows; the resulting Excel export was parsed and all 18 expected headline values matched to four decimal places.

## Recipient experience now

3 actions, no SPSS knowledge required:

1. Extract the zip to any folder
2. Double-click `0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat` (Windows) or `0_DOUBLE_CLICK_ME_TO_START_MAC.command` (macOS, right-click → Open the first time for Gatekeeper)
3. The launcher pre-installs `semopy` (first run ~30 sec), then SPSS opens with the syntax preloaded; click `Run > All`

After ~90 seconds, the same folder contains:

- `spv_export.xlsx` -- all SPSS tables in Excel format (open in Excel/Numbers/LibreOffice)
- `Post_Run_Results.spv` -- native SPSS Viewer document

SPSS stays open afterwards so the recipient can drill into individual tables, save additional outputs, or re-run blocks interactively.

## What changed

- **`scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`** (new) -- locates SPSS by searching common install paths newest-first; pre-installs `semopy==2.3.11` via the SPSS-bundled Python; generates a combined runtime syntax file (`_run_validation.sps`) in the extraction folder containing an absolute `CD '<folder>'.` followed by the verbatim contents of the main syntax; launches SPSS with that combined file. Falls back to printing manual instructions if SPSS is not found.
- **`scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`** (new, executable bit set in git) -- macOS counterpart using the same combined-file approach, launched via `open -a`.
- **`scripts/spss/tabs_v2_validation.sps`** -- substantially expanded beyond the original small edits: removed the broken `MVA /CATEGORICAL=NONE` subcommand; added `OUTPUT EXPORT /XLSX` and `OUTPUT SAVE`; added new sections for CFA via embedded Python (`semopy`) and R (`lavaan`/`semTools`) including measurement invariance tests; added Mardia multivariate normality tests. The syntax now covers 15 sections for a full psychometric walkthrough.
- **`scripts/cross-platform-verification/build_zip.py`** -- SPSS_MANIFEST now includes both launchers; `_add_to_zip` helper sets a fixed `0o100755` (rwxr-xr-x regular-file) mode on `.command` and `.sh` files when zipping, with `create_system=3` (Unix) forced so macOS unzippers honour the permission bits even when the zip is built on Windows; SPSS_README inside the zip rewritten to lead with the double-click flow.
- **`scripts/spss/README.md`** -- easiest-workflow section rewritten to lead with the launcher; file table updated.
- **`scripts/cross-platform-verification/tabs_v2_validation_spss.zip`** -- regenerated with the new launchers.

## Why a combined file (rationale captured in the .bat header comment)

Two SPSS launch behaviors broke earlier iterations:

1. `stats.exe FILE.sps` sets SPSS's working directory to the install folder (`C:\Program Files\IBM\SPSS Statistics\`), not where the .sps lives. Relative paths in the syntax then resolve incorrectly. Earlier attempts at using `start /D` did not consistently override this.
2. `INSERT FILE='...'` routes the inserted commands' output to a sub-document in some SPSS versions, leaving the main Viewer empty when `OUTPUT EXPORT` runs at the end.

Concatenating an absolute `CD` with the main syntax content into a single file avoids both. The combined file is regenerated every time the launcher runs and is safe to delete after the validation completes.

## End-to-end test results (Windows + SPSS 31.0)

All 18 documented headline values reproduced to four decimal places via the double-click flow:

| Metric | Auto-Run | Expected |
|---|---|---|
| Cronbach alpha (B/R/M) | 0.8728 / 0.9171 / 0.8847 | 0.873 / 0.917 / 0.885 |
| Listwise N (B/R/M) | 192 / 181 / 191 | 192 / 181 / 191 |
| KMO (B/R/M) | 0.8508 / 0.9266 / 0.9121 | 0.851 / 0.927 / 0.912 |
| Bartlett chi-sq (B/R/M) | 1135.5105 / 1278.9893 / 650.1338 | 1135.51 / 1278.99 / 650.13 |
| Pearson r (B-R / B-M / R-M) | -0.401 / -0.333 / +0.732 | ~ -0.40 / -0.33 / +0.73 |

Bit-identical to a hand-run of the same syntax through the SPSS GUI.

## Test plan

- [x] Author tested double-click flow end-to-end on Windows + SPSS 31.0
- [x] Output `spv_export.xlsx` parsed; all 18 headline values match expected
- [x] Output `Post_Run_Results.spv` produced at correct size (~147 KB)
- [x] SPSS stays open after the run (Viewer + data + syntax all visible)
- [x] Build script regenerates the SPSS zip cleanly with the new launchers
- [x] `.command` file's executable bit preserved through the zip via `_add_to_zip`
- [ ] Reviewer to spot-check the combined-file approach in the .bat / .command
- [ ] Reviewer to confirm the file-table and Easiest-workflow sections in `README.md` read clearly

## Follow-ups (not in this PR)

- Apply the same double-click pattern to the Minitab bundle (`scripts/minitab/`) once we have a tested approach for Minitab Exec macros launched headlessly
- Consider adding a CI smoke that builds the zip and asserts the .bat / .command files are present and the .command has the executable bit set


clarkemoyer:

1. `scripts/spss/README.md`: The docs claim the launcher runs SPSS “headlessly” with “no manual clicks” and that the user will see “DONE”, but the launcher scripts open SPSS interactively and require the recipient to click Run > All; they also never print DONE. Please update these steps to match the actual flow (or implement DONE/status reporting if that’s intended).
2. `scripts/spss/README.md`: This file table entry says the Windows launcher “runs the syntax in production mode…prints Done”, but Run_Validation.bat only opens SPSS with the combined syntax and does not run it or print a DONE signal. Please reword to avoid implying unattended execution/completion signaling (or add that behavior).
3. `scripts/cross-platform-verification/build_zip.py`: The bundled README text says the launcher runs with “no GUI clicks required” and instructs users to wait for “DONE”, but the launchers require the user to click Run > All and never emit DONE. Please align SPSS_README’s quick-start wording with the actual interactive flow.
4. `scripts/cross-platform-verification/build_zip.py`: SPSS_README says `tabs_v2_validation.sps` “auto-runs from the launcher”, but the launcher scripts only open SPSS with the file preloaded; the user must still click Run > All. Please adjust this description to avoid overpromising automation.
5. `scripts/spss/tabs_v2_validation.sps`: The comment claims the launchers “pop a Done dialog”, but neither launcher script nor this syntax file triggers a dialog/notification when the run finishes. Please remove/adjust this claim or implement an explicit completion signal.

6. `scripts/spss/Run_Validation.command`: The generated SPSS command `CD '$HERE'.` will break if the extracted folder path contains a single quote ('). SPSS expects embedded quotes to be escaped (typically doubled). Please escape/sanitize $HERE when writing the combined syntax file so it works for paths containing apostrophes.
7. `scripts/cross-platform-verification/build_zip.py`: When this script is run on Windows, `ZipInfo.from_file` defaults `create_system` to MS-DOS; some unzip tools then ignore POSIX permission bits in `external_attr`, so the `.command` may still lose its executable bit. Consider forcing `info.create_system = 3` (Unix) when setting `external_attr` to reliably preserve +x across build platforms.

8. `scripts/spss/Run_Validation.bat`: The generated SPSS command `CD '%HERE_FWD%'.` will break if the extracted folder path contains a single quote ('). Please escape/sanitize the path when writing the combined syntax file (SPSS typically requires doubling embedded quotes) so the launcher works from directories containing apostrophes.
9. `scripts/spss/Run_Validation.command`: The script header says it “requires SPSS Statistics 24+”, but the app search list only checks unversioned + 27–31 paths. If a recipient has SPSS 24–26 installed under the common “IBM SPSS Statistics <version>” folder name, this launcher won’t find it. Consider adding 24–26 candidates (or a more general glob-based discovery).


clarkemoyer:
**Copilot Review Cycle (Round 1/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/spss/Run_Validation.command`: In the single-quote escaping logic, `HERE_ESCAPED="${HERE//\'/\'\'}"` won’t actually replace `'` characters because inside double quotes the backslashes are treated literally (so the pattern becomes `\'`). This will generate an invalid `CD '...'.` line if the extracted folder path contains a single quote.

Consider changing the substitution to one that matches literal `'` and replaces it with `''` (two single quotes) so SPSS parses the path correctly.
2. `scripts/cross-platform-verification/build_zip.py`: The bundled zip README now claims this is “Built for IBM SPSS Statistics 24+…”, but the in-repo SPSS README still states it was built/tested against SPSS Statistics 31.0. This mismatch is likely to confuse recipients about the supported versions.

Suggestion: align the stated minimum/tested SPSS version across both READMEs (e.g., “tested on 31.0; expected to work on 24+” if that’s the intent).

clarkemoyer:

1. `scripts/spss/Run_Validation.bat`: The SPSS discovery list for `%ProgramFiles(x86)%` only includes versions 31/30/29, but the script claims support for 24+. On 32-bit installs (or where SPSS is installed under Program Files (x86)), versions 24–28 won’t be found. Consider adding the missing versioned paths (28..24) under `%ProgramFiles(x86)%` (and/or broadening the search strategy) so the launcher works for the stated range.

2. `scripts/spss/Run_Validation.bat`: The search order checks the unversioned path `%ProgramFiles%\IBM\SPSS Statistics\stats.exe` before the version-specific directories (31→24). If multiple SPSS versions are installed, this can select an older install even though the intended behavior is “newest-first”. Consider moving the unversioned path to the end (as a fallback), or explicitly preferring the highest detected version.
3. `scripts/spss/Run_Validation.command`: The SPSS app search checks the unversioned "/Applications/IBM SPSS Statistics/SPSSStatistics.app" before the versioned folders (31→24). If both exist, this may pick an older install even though the launcher is intended to prefer the newest available version. Consider reordering the candidates to check the highest version first and only fall back to the unversioned path if no versioned app bundles are present.
4. `scripts/spss/Run_Validation.command`: The script prints a success message after calling `open -a ...` but does not verify that `open` succeeded. If `open` fails (e.g., app bundle not launchable, Gatekeeper/quarantine, bad association), the user will still be told “SPSS is opening”. Consider checking the exit status of `open` and printing actionable instructions (and exiting non-zero) when it fails; alternatively enable `set -e` and handle the failure path explicitly.


clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The comment here implies that `CD '.'.` makes the working directory “the same folder this syntax file is in”. In SPSS, `CD '.'` only sets the working directory to the current working directory at runtime; it does not reliably resolve relative to the syntax file’s location (and this is why the launcher prepends an absolute `CD`). Suggest rewording to say outputs go to the current working directory, and note that the launcher/absolute `CD` ensures that is the extracted folder.
2. `scripts/cross-platform-verification/build_zip.py`: In the manual fallback section, this says the syntax writes outputs “into the folder where the .sps file lives.” That’s not necessarily true if SPSS’s working directory isn’t that folder (the syntax uses `CD '.'`, which is relative to the current working directory). Consider changing this wording to “into the current working directory” and explicitly instruct the user to set it to the unzipped folder (which you already mention below).


clarkemoyer:

1. `scripts/spss/Run_Validation.bat`: `enabledelayedexpansion` doesn't appear to be used anywhere in this script, and leaving it on can cause unexpected behavior if the extracted folder path contains `!` (it can be interpreted by CMD's delayed-expansion parser). Consider removing `enabledelayedexpansion` (or disabling delayed expansion around any lines that work with paths) to make the launcher more robust for arbitrary extraction locations.

2. `scripts/spss/Run_Validation.bat`: The `echo CD '%HERE_FWD_ESC%'.` line is vulnerable to CMD metacharacters in the extraction path (e.g., `&`, `|`, `<`, `>`), which can break the command (or execute unintended extra commands) and produce a corrupted `_run_validation.sps`. Please escape CMD metacharacters in the path before echoing it into the file, or write the file using a safer mechanism that doesn't re-parse special characters.
3. `scripts/spss/tabs_v2_validation.sps`: The comment says the launcher script can "confirm the run finished", but neither `Run_Validation.bat` nor `Run_Validation.command` currently checks for the presence of `Post_Run_Results.spv` (or `spv_export.xlsx`)—they only open SPSS. Please adjust this comment to match the actual behavior, or add an explicit post-run check if that confirmation is intended.


clarkemoyer:

1. `scripts/spss/Run_Validation.bat`: `%~dp0` returns a trailing backslash; stripping it unconditionally breaks the edge case where the zip is extracted to the root of a drive (e.g., `C:\`). In that case `HERE` becomes `C:` (not `C:\`), which can cause `cd /d` and the generated `CD '...'.` line to point at the wrong directory. Consider keeping the trailing slash, or only trimming it when the path length is > 3 / not a drive root, or using `pushd "%~dp0"` to avoid this entirely.


clarkemoyer:

clarkemoyer:

1. `scripts/spss/README.md`: This workflow description says SPSS “runs headlessly in production mode (no manual clicks)”, but both launchers actually open SPSS interactively and instruct the user to click Run > All. Please update the wording to reflect the real interaction (or, if you intended true headless execution, change the launchers to invoke SPSS in production mode and run the syntax automatically).

2. `scripts/cross-platform-verification/build_zip.py`: The embedded SPSS bundle README claims the launcher “runs the syntax in production mode (no GUI clicks required)” and that recipients should wait for bootstrap blocks and a DONE message. However, the launchers in this PR open SPSS interactively and require the user to click Run > All, so these instructions are currently inaccurate. Align this README with the actual launcher behavior (or update the launcher to truly run headless).
3. `scripts/cross-platform-verification/build_zip.py`: `_add_to_zip` sets `external_attr` for `.command`/`.sh` files, but it doesn’t force `ZipInfo.create_system=3` (Unix). `ZipInfo.from_file()` sets `create_system` based on the OS that builds the zip, so regenerating the bundle on Windows can produce a zip where macOS tools ignore the Unix permission bits and the `.command` extracts non-executable. Consider explicitly setting `info.create_system = 3` for executable entries to make the zip deterministic across build platforms.
4. `scripts/cross-platform-verification/build_zip.py`: The PR description references new launchers named `Run_Validation.bat` / `Run_Validation.command`, but this change set adds `0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat` and `0_DOUBLE_CLICK_ME_TO_START_MAC.command` (and the zip/manifest uses those names). Please update the PR description (or rename the files) so the documented filenames match what’s actually shipped.
5. `scripts/cross-platform-verification/build_zip.py`: The SPSS bundle README’s “What’s inside” list says `tabs_v2_validation.sps` “auto-runs from the launcher”, but the launchers in this PR open SPSS with the syntax preloaded and require the user to click Run > All. Update this line to avoid implying the syntax executes automatically.

6. `scripts/spss/README.md`: The README tells recipients to look for a "DONE" message before expecting output files, but neither launcher currently prints a DONE marker (the Windows batch script pauses unconditionally after launching SPSS, and the macOS script waits for Enter). Unless SPSS itself prints an explicit DONE message somewhere, this instruction will confuse recipients; either add an actual DONE signal in the launcher, or change the README to describe how to confirm completion (e.g., wait for `spv_export.xlsx` / `Post_Run_Results.spv` to appear).
7. `scripts/spss/tabs_v2_validation.sps`: The normality section comment says multivariate normality is in “Section 16”, but this file is structured as 15 sections and the Mardia multivariate normality code is in Section 13. This incorrect section reference makes the walkthrough harder to follow; update the comment to point to the correct section.

8. `scripts/spss/README.md`: The file table and workflow instructions refer to `0_DOUBLE_CLICK_ME_TO_START.bat` / `0_DOUBLE_CLICK_ME_TO_START.command`, but the launcher files committed here are `0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat` and `0_DOUBLE_CLICK_ME_TO_START_MAC.command`. Update the README to use the actual filenames so recipients can follow the instructions successfully.
9. `scripts/spss/tabs_v2_validation.sps`: The header comment says “Sections 14-17 of this syntax bridge that gap”, but the file is organized as 15 sections and the SEM bridge appears to be Sections 13–14. Update the section range in the comment so it matches the actual structure.
10. `scripts/spss/tabs_v2_validation.sps`: The SPSS syntax header references launcher filenames that don't exist in this repo/zip (`0_DOUBLE_CLICK_ME_TO_START.bat` and `0_DOUBLE_CLICK_ME_TO_START.command`). The actual files are `0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat` and `0_DOUBLE_CLICK_ME_TO_START_MAC.command`, so recipients following these instructions will fail to find the right file. Update the header text to match the real filenames (and keep it consistent with the bundle README).

clarkemoyer:
**Copilot Review Cycle (Round 1/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: When writing the `CD '%HERE_FWD%'.` line into `_run_validation.sps`, `%HERE_FWD%` is not escaped for cmd.exe metacharacters (notably `&`, `|`, `<`, `>`, `^`). If the extracted folder path contains any of these (e.g., `R&D`), the `echo` command will be parsed incorrectly and the combined syntax file will be corrupted. Escape the path before echoing it (and also handle single quotes for SPSS string literal safety) or generate the file via a safer writer (e.g., PowerShell/Python).

2. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: The launcher installs `semopy` without pinning a version. For reproducibility (and to match the canonical pipeline), consider installing the pinned version from `scripts/analysis/requirements-validation.txt` (currently `semopy==2.3.11`) instead of the latest PyPI version.

3. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: The combined syntax file writes `CD '$HERE'.` using single quotes. If the extracted folder path contains an apostrophe, SPSS will see an unterminated string and fail to set the working directory. Consider escaping single quotes for SPSS syntax (or using a different quoting/encoding strategy) when emitting the CD command.
4. `scripts/cross-platform-verification/build_zip.py`: The PR description refers to launcher filenames like `Run_Validation.bat` / `.command`, but the implementation and bundle manifest use `0_DOUBLE_CLICK_ME_TO_START_*`. Please align the PR description (and any external instructions) with the actual shipped filenames, or rename the files if the `Run_Validation.*` names are required.
5. `scripts/spss/tabs_v2_validation.sps`: The PR description indicates `tabs_v2_validation.sps` only had small edits (MVA portability + OUTPUT EXPORT/SAVE), but this file now includes substantial additional flow/sections and embedded Python/R SEM blocks. If that scope increase is intentional, please update the PR description to reflect it; otherwise, consider splitting the large syntax refactor/feature-add into a separate PR to keep the change reviewable.
6. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: The launcher installs `semopy` without pinning a version. For reproducibility (and to match the canonical pipeline), consider installing the pinned version from `scripts/analysis/requirements-validation.txt` (currently `semopy==2.3.11`) instead of whatever happens to be latest on PyPI at runtime.

clarkemoyer:

1. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: The launcher tells recipients to press Ctrl+C to skip the R package install. In cmd.exe, Ctrl+C typically aborts the entire batch file (prompting to terminate the batch job), so this can prevent SPSS from launching at all. Replace this with an explicit choice (e.g., prompt Y/N to install R packages) that cleanly skips the R step without terminating the script.

2. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: The single-quote escaping for the generated SPSS `CD '...'.` line is incorrect: `HERE_SPS="${HERE//\'/\'\'}"` searches for a backslash+quote sequence (because `\'` is literal inside double quotes), so paths containing `'` won’t be escaped and will break the SPSS syntax. Replace this with a true `'` → `''` substitution (no backslashes) so apostrophes in folder names are handled safely.

3. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: The instructions to “Press Ctrl+C now to skip” the R package install are misleading: Ctrl+C will typically terminate the entire launcher script (not just skip the R step), leaving the user without launching SPSS. Prefer an explicit prompt (Y/N) to install lavaan/semTools, or a command-line flag, so skipping cleanly continues to the SPSS launch.


clarkemoyer:

1. `scripts/cross-platform-verification/build_zip.py`: SPSS_README’s “Targeted SPSS license” section still says that without Amos the bundle covers only the descriptive/reliability layer and that CFA/HTMT/bifactor/invariance/Mardia stay in the Python pipeline. However, tabs_v2_validation.sps now computes many of these via embedded Python/R (Sections 13–14). Please update this text to accurately reflect the bundle’s current outputs and prerequisites (semopy/lavaan installs), or gate those sections if they’re intentionally out of scope for this zip.
2. `scripts/cross-platform-verification/build_zip.py`: This Quick start estimates “~60 seconds”, but the Windows/macOS launchers currently tell recipients to expect about 90 seconds. Please standardize the runtime estimate across this README and the launcher output so recipients don’t assume the run is stuck.

3. `scripts/spss/README.md`: This README says “After ~60 seconds” results appear, but the Windows/macOS launchers prompt recipients to expect about 90 seconds. Please standardize the runtime estimate across the README and the launcher output to reduce confusion for recipients.

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: Section 8’s header comment claims Cohen’s d is computed “below … via COMPUTE”, and the section lists `cohens_d_smb_ent` as an output, but there is no subsequent SPSS syntax that actually computes or prints Cohen’s d (the only `COMPUTE` later is `row_index`). Either add the Cohen’s d computation/output or remove the claim so recipients aren’t expecting an effect-size table that never appears.

2. `scripts/spss/README.md`: The README row for `tabs_v2_validation.sps` says it “runs every analysis the targeted license supports natively”, but the syntax file now includes embedded Python/R sections (13–14) that go beyond native SPSS procedures. Consider rewording this description to avoid implying the whole workflow is SPSS-native, or explicitly call out the Python/R extensions here as well.

3. `scripts/cross-platform-verification/build_zip.py`: When building on Windows, `src_path.stat().st_mode` commonly yields permissions equivalent to 0666; OR-ing with `0o100111` then writes `0o100777` into `external_attr`, making the extracted `.command`/`.sh` world-writable on macOS. To keep permissions minimal and predictable, consider explicitly forcing 0o100755 (or stripping group/other write bits) before setting `external_attr`, while still ensuring +x and `create_system=3`.

4. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: This banner says it is checking “numpy, pandas, scipy, semopy”, but the actual probe only runs `import semopy`. If numpy/pandas/scipy are missing from the SPSS-bundled Python, Section 13 will still fail later. Either expand the probe to import all four (and print clearer guidance) or change the banner text to match what’s being checked.

5. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: This banner says it is checking “numpy, pandas, scipy, semopy”, but the actual probe only runs `import semopy`. If numpy/pandas/scipy are missing from the SPSS-bundled Python, Section 13 will still fail later. Either expand the probe to import all four (and print clearer guidance) or change the banner text to match what’s being checked.


clarkemoyer:
**Copilot Review Cycle (Round 5/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/cross-platform-verification/build_zip.py`: In `_add_to_zip`, the comment says permissions are kept at `0o100755` regardless of build platform, but `exec_mode = (src_mode | 0o100111) & ~0o022` can still preserve a more restrictive mode (e.g., `0o100700`) if the source file is owner-only. If the intent is always `755` inside the zip, consider explicitly setting `info.external_attr = 0o100755 << 16` (or at least OR in group/other read+execute bits) rather than deriving from `stat()`.

2. `scripts/cross-platform-verification/build_zip.py`: `SPSS_README`’s “What’s inside” code block has inconsistent spacing/alignment (the macOS launcher line is shorter than the Windows one), which makes the bundle’s top-level README harder to scan. Consider aligning the columns so the launcher filenames and descriptions line up cleanly.


clarkemoyer:

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The Section 13 manual-install instructions hardcode a Windows path ("C:\Program Files\IBM\SPSS Statistics\Python3\python.exe") that likely won't exist for versioned installs (e.g. the launcher searches under "...\IBM\SPSS\Statistics\31\Python3\python.exe"). Consider rewriting this to reference the discovered SPSS install folder (or list both common layouts) so recipients don’t follow a non-working command.

2. `scripts/spss/tabs_v2_validation.sps`: Section 13 currently sets `SEM_OK = False` when `semopy` is missing, which skips *all* of Section 13 (including HTMT and Mardia) even though those computations only require numpy/pandas/scipy. Consider splitting this into separate flags (e.g., core_ok vs semopy_ok) so non-semopy outputs can still run when semopy install/import fails.

3. `scripts/spss/tabs_v2_validation.sps`: The Section 14 header comment says this block provides “Delta-CFI / Delta-RMSEA” invariance tests, but the code only prints Delta-CFI. Either compute/report Delta-RMSEA as documented or update the comment to avoid implying outputs that aren’t produced.

4. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: The generated combined syntax header says it “INSERTs tabs_v2_validation.sps”, but the launcher actually concatenates the file contents via `type ... >>` (no SPSS `INSERT FILE=` command). Consider adjusting the wording so the console/log output matches what’s happening (helps debugging when recipients report issues).
5. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: Similarly to the Windows launcher, the combined syntax header says it “INSERTs tabs_v2_validation.sps”, but the script is actually concatenating the file contents with `cat`. Consider updating the wording to reflect concatenation rather than SPSS `INSERT FILE=`.
6. `scripts/spss/README.md`: In the file table you describe `tabs_v2_validation.sps` as computing CFA/SEM outputs via embedded Python/R (Sections 13–14). Earlier in this README it also states that without Amos “SPSS cannot natively compute” CFA/omega/CR/AVE/etc and recommends citing the Python pipeline for those. Please reconcile the wording to clearly distinguish “not native” from “this bundle bridges via embedded Python/R”, so recipients don’t get contradictory guidance about what SPSS will produce.

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The skip note says “HTMT, HTMT2, and Mardia will still run (they need only numpy/scipy)”, but this block doesn’t import/check `scipy` unless semopy was installed (and semopy is what typically pulls in scipy). If scipy is missing, Mardia will fail and the note becomes misleading. Consider either importing/validating scipy in the `core_ok` import block (and adjusting the launcher check accordingly), or updating the message to reflect that Mardia also depends on scipy being present.
2. `scripts/spss/tabs_v2_validation.sps`: In the R fallback, `df <- spssdata.GetDataFromSPSS(...)` is executed without any guard/tryCatch. If the SPSS R integration doesn’t expose `spssdata` in this environment (or data extraction fails), the whole Section 14 block will error out despite lavaan being installed. Wrap the dataset fetch in `tryCatch` (or check for the accessor’s existence first) and print a clean “[skip] …” message so the overall validation run can continue.


clarkemoyer:

1. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: The Windows launcher’s SPSS install-path search list uses the `IBM\SPSS\Statistics\<version>` pattern, but this PR’s own manual install-path examples (in `tabs_v2_validation.sps`) use `IBM\SPSS Statistics\<version>` (space, no extra subfolder). To avoid false “SPSS not found” failures on common installs, include the `IBM\SPSS Statistics\<version>` candidates (and x86 equivalents) in the search order, matching the paths documented in the syntax comments.
2. `scripts/spss/README.md`: The block-quoted “Recommended dissertation paragraph” is currently broken by an empty quoted line followed by a quoted bullet (`> - Bootstrapping...`), which renders as a list and disrupts the sentence. Reformat this quote as continuous prose (or a properly formatted list) so it reads correctly in Markdown.

3. `scripts/cross-platform-verification/build_zip.py`: In the generated SPSS bundle README text, it says the launcher “pre-installs automatically” both the embedded Python and embedded R dependencies. However, the launchers prompt before installing R packages (lavaan/semTools) and may skip them. Consider rewording this section to distinguish: semopy is auto-installed, while R packages are optional and user-confirmed.


clarkemoyer:

1. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: `%HERE%` is echoed and piped into `findstr` without escaping CMD metacharacters. If the extracted folder path contains characters like `&`, these lines can break (e.g., `echo Working folder: %HERE%` will treat `&` as a command separator), causing the launcher to fail even though SPSS is installed. Consider escaping/quoting for display + `findstr` (e.g., precompute an escaped path for echo/piping) so the double-click flow works from any normal Windows folder name.
2. `scripts/cross-platform-verification/build_zip.py`: PR description says the `.command`/`.sh` executable bit in the zip is derived from the source file’s `st_mode`, but the implementation here forces a fixed `0o100755` for all `.command`/`.sh` entries. Either update the PR description/docs to match this fixed-mode behavior, or change the code if preserving the source mode is actually required.
3. `scripts/cross-platform-verification/build_zip.py`: This script rebuilds and commits *all* bundles (`tabs_v2_validation_spss.zip`, `tabs_v2_validation_minitab.zip`, `tabs_v2_validation_python.zip`). In this PR, `tabs_v2_validation_minitab.zip` was also updated, but the PR description focuses on the SPSS bundle. If the Minitab zip change is intentional (e.g., as a side-effect of the new `_add_to_zip` behavior), please call that out in the PR description; otherwise consider reverting the Minitab zip regen to keep the PR scope tight.
4. `scripts/spss/tabs_v2_validation.sps`: Section 10 uses `row_index` as the regression dependent, which is fine for Mahalanobis/leverage (they depend only on predictors) but makes **Cook’s distance** (and any residual/DFBETA-style influence diagnostics) meaningless because they depend on residuals from the DV model. Also, the header comment claims “Standardized DfBeta per case (saved as SDB_*)”, but the `/SAVE` subcommand here doesn’t request any DFBETA/SDBETA outputs. Consider either (a) saving only MAHAL/LEVER here, or (b) using a meaningful DV if you want Cook’s D / influence metrics, and align the header comment with what’s actually saved.
5. `scripts/spss/README.md`: The README’s license/module table says **Advanced Statistics** is “not required by the syntax in this folder”, but `tabs_v2_validation.sps` now runs a `GLM ... BY SMB_ENT` MANOVA block and the syntax itself labels that section as “Advanced Statistics”. Please update the table (and/or mark the MANOVA block as optional) so the documented required modules match what the syntax actually executes.


clarkemoyer:

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The manual semopy installation commands in this prerequisites block install an unpinned `semopy` version. The launchers pre-install `semopy==2.3.11`, and the expected-value comparisons assume a consistent semopy version. Pin the manual install instructions to the same version (and ideally mirror the launcher command) so manual runs are reproducible.

2. `scripts/spss/tabs_v2_validation.sps`: This runtime note prints a manual install command for semopy without pinning the version. Since the launcher pins `semopy==2.3.11`, printing an unpinned install command can lead to recipients installing a newer semopy and getting different fit indices/omega values. Update the message to instruct installing the pinned version to match the launcher.

3. `scripts/spss/README.md`: The README still claims the Bootstrapping module is used to compute bootstrap percentile CIs around Cronbach’s alpha, and later refers to running RELIABILITY inside a BOOTSTRAP wrapper. In the updated `tabs_v2_validation.sps`, the BOOTSTRAP wrappers around RELIABILITY were removed (and the syntax notes that alpha bootstrap support is version/license-dependent), so this documentation is now inaccurate. Update this module table (and the later analysis table entries) to describe what bootstrapping is actually used for (e.g., correlation/regression CIs) and/or explicitly state that alpha CIs are only computed in the canonical Python pipeline.

clarkemoyer:
**Copilot Review Cycle (Round 5/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/spss/README.md`: The Mahalanobis-outlier rows in this table are now out of sync with the updated syntax. In `tabs_v2_validation.sps` Section 10, Mahalanobis is computed via `REGRESSION` with `row_index` as the dependent variable and `/SAVE MAHAL LEVER`, not with `B_mean`/`R_mean`/`M_mean` as the dependent variable. Please update the commands shown here so recipients can reconcile the table with the actual syntax/output variables (MAH_1/LEV_1 etc.).
2. `scripts/cross-platform-verification/build_zip.py`: The SPSS bundle README text still describes the bundle as reproducing only the “descriptive + reliability layer”, but the bundled `tabs_v2_validation.sps` now includes Sections 13–14 that run CFA/SEM-derived statistics via embedded Python/R. Consider updating this opening description to reflect the expanded scope so recipients aren’t surprised by the semopy/lavaan installation prompts and SEM outputs.

3. `scripts/cross-platform-verification/build_zip.py`: PR description says the `.command` executable-bit preservation is derived from the source file’s `st_mode | 0o100111`, but the implementation here forces a fixed `0o100755` for `.command`/`.sh`. Either approach can be valid, but the PR description (or the code comments) should be updated to match the actual behavior to avoid confusion during future maintenance/regenerations.

clarkemoyer:

clarkemoyer:

1. `scripts/spss/tabs_v2_validation.sps`: The prerequisites comment says semopy must be installed in the SPSS-bundled Python “site-packages”, but the documented commands (and the launcher) install with `pip install --user`, which targets the user-site directory. To avoid confusing recipients troubleshooting imports, please reword this to explicitly say “user site-packages” (or remove the site-packages wording and just point to the exact command).

clarkemoyer:
**Copilot Review Cycle (Round 1/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:

1. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_WINDOWS.bat`: The semopy pre-install check only verifies that `import semopy` succeeds; if semopy is already present but at a different version than the pinned `2.3.11`, the launcher will skip installation and the validation may run with an unintended semopy version (hurting reproducibility vs the canonical pipeline). Consider checking `semopy.__version__` (and re-installing/forcing `semopy==2.3.11` when it differs) rather than only checking import success.

2. `scripts/spss/0_DOUBLE_CLICK_ME_TO_START_MAC.command`: The semopy install guard only checks whether `import semopy` works; if semopy is installed but not at the pinned `2.3.11`, this will silently keep the wrong version and may change CFA/SEM results. Consider checking `semopy.__version__` (and forcing a reinstall/downgrade to `2.3.11` when needed) instead of treating any installed semopy as acceptable.

3. `scripts/cross-platform-verification/build_zip.py`: PR description mentions preserving executable bits derived from the source file’s `st_mode`, but `_add_to_zip()` now forces a fixed `0o100755` mode for `.command`/`.sh` regardless of the source permissions. Either update the PR description to match the implementation, or (if deriving from source mode is important) adjust `_add_to_zip()` to compute the mode from `src_path.stat().st_mode` in a controlled way.
4. `scripts/spss/tabs_v2_validation.sps`: The Section 9 regressions specify `/NOORIGIN`, which forces the models through the origin (no intercept). That materially changes coefficients/R² and is nonstandard for these construct-mean regressions; it also isn’t mentioned in the section header text. If the intent is to mirror the canonical Python/R regressions, consider removing `/NOORIGIN` for Section 9 (keeping it only where it’s truly required, e.g., the Mahalanobis proxy-regressions in Section 10).

clarkemoyer:

PR #1873, opened 2026-05-03
refactor(results): extract shared DescriptiveContent component (live + CRP-2026 parity)

clarkemoyer:
## Summary

Sixth refactor in the live ↔ CRP-2026 shared-components series, following PRs #1866 (top-barriers), #1867 (data-quality), #1870 (sensitivity, merged), #1871 (findings, in flight), and #1872 (reliability, in flight). Consolidates the live and CRP-2026 descriptive-statistics pages into a single shared `<DescriptiveContent variant="live"|"crp" data={...}/>` component.

## Variant differences captured in `VARIANT_CONFIG`

- **Title and date treatment** — `<LastUpdated>` for live, static green "Published: April 2026" badge for CRP
- **Sample filtering** — CRP shows 3 nested groups; live shows all 5 for the construct tables but only 4 for the correlation matrices (matching the original live page's hardcoded 4-group correlation list, preserved via `correlationExcludeKeys: ['v2_all']`)
- **Intro values clause** — "from the daily analysis pipeline" vs "from the frozen CRP 2026 dataset across three sample definitions"
- **Correlation intro count word** — "four primary result groups" vs "three CRP sample groups"
- **Correlation footnote** — live → `/results/sensitivity` for the full five-sample comparison; CRP → `/results/crp-2026` for dataset methodology
- **Related-page link sets** — live links Sensitivity Analysis; CRP links the CRP 2026 Overview
- **Back-link target** — `/results/crp-2026` for CRP (matches the PR #1872 fix), `/results` for live

## A11y improvements

`scope="col"` / `scope="row"` added on all four tables (both originals were missing these).

## Files

- `src/components/results/descriptive/DescriptiveContent.tsx` (new, shared, with JSDoc)
- `src/app/results/descriptive/page.tsx` (309 → 30 lines, with `SEARCH_CONTENT` seed)
- `src/app/results/crp-2026/descriptive/page.tsx` (314 → 28 lines, with `SEARCH_CONTENT` seed)
- `__tests__/components/DescriptiveContent.test.tsx` (21 tests, jest-axe a11y)
- 4 already-merged shells (top-barriers + data-quality, both live + CRP): `SEARCH_CONTENT` seeds added — no-op if PR #1871 or #1872 lands first
- `public/search-index.json` removed from tracking (in `.gitignore` line 114, regenerated by `prebuild`; matches the cleanup on PR #1872)

## Test plan

- [x] `npm run format` — clean
- [x] `npm run lint` — 0 errors, 3 expected warnings
[...]

PR #1881, opened 2026-05-04
feat(ci): phantom-revert guard for stale-base PRs

clarkemoyer:
## Summary

We hit the same stale-base bug twice on 2026-05-03:

- **#1874** (Minitab launcher) -- branch forked before #1873/#1860 merged; would have reverted committed binary zips.
- **#1876** (pipeline-gap stats) -- branch forked before #1873/#1860/#1874 merged; would have reverted shared content components, SPSS launchers, and Minitab launchers.

Each time we caught it during manual pre-merge diff review. This guard catches it automatically.

## How it works

- Computes merge-base between PR head and base ref.
- Lists files modified on base since the fork point that the PR does not touch.
- Hard-fails when the branch is >5 commits behind OR any phantom-revert candidate falls in a critical path: `src/data/`, `src/components/`, `scripts/analysis/`, `scripts/cross-platform-verification/`, `scripts/spss/`, `scripts/minitab/`, `.github/workflows/`.
- Always writes a categorized report to the workflow step summary.

## Defense in depth

Pair with GitHub branch protection's **"Require branches to be up to date before merging"** rule:

| Layer | Role |
|---|---|
| Branch protection (built-in) | Forces fresh-CI-after-update pass before merge button enables |
| This guard (custom) | Explains *why* the update is needed when there are at-risk files |

## Test plan

- [ ] CI green on this PR (the guard should report 0 phantom candidates against fresh main)
- [ ] After merge: open a deliberately-stale test branch (forked from old main), verify the guard fails with a clear report

### DIFF
--- a/.github/workflows/phantom-revert-guard.yml
+++ b/.github/workflows/phantom-revert-guard.yml
@@ -5,23 +5,27 @@ name: Phantom Revert Guard
 #
 #   1. Branch B is forked from main at commit M0.
 #   2. PR P1, P2, ... merge into main between M0 and now (HEAD = M_now).
-#   3. Branch B's CI passes because it's still building against M0's tree.
-#   4. The PR's "diff vs main" looks like (B - M_now), which removes everything
-#      P1, P2, ... added in files that B doesn't touch -- a phantom revert.
+#   3. Branch B is not updated; CI still passes because it tests B merged with
+#      M_now (the virtual merge commit), which is fine as a build -- but a
+#      subsequent squash- or rebase-merge of the stale branch writes B's old
+#      file tree, silently reverting everything P1, P2, ... added to files B
+#      never touched.
+#   4. The PR's "diff vs main" shows those files as deleted/reverted.
 #
-# A regular CI run is happy because it builds B in isolation. This guard
-# explicitly compares "files modified on main since fork" against "files
-# changed by this PR" and fails if the PR appears to drop content from any
-# file that main has updated.
+# This guard explicitly compares "files modified on main since fork" against
+# "files changed by this PR" and fails when untouched-but-updated files fall
+# in critical paths or the branch is sufficiently far behind.
 #
-# Pair this guard with branch protection's "Require branches to be up to date
-# before merging". The branch-protection rule forces the PR to merge or rebase
-# main in before the merge button enables; this guard explains *why* if the
-# attempted merge would produce a phantom revert.
+# Note: this guard reruns only when a new commit is pushed to the PR branch.
+# If main advances while a PR is open and the PR author pushes nothing new,
+# the guard will not automatically rerun. For full coverage, pair this guard
+# with branch protection's "Require branches to be up to date before merging"
+# rule, which forces a fresh CI pass after updating the branch.
 
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+  workflow_dispatch: # Allow manual reruns when main has advanced
 
 permissions:
   contents: read
@@ -54,6 +58,9 @@ jobs:
         run: |
           set -euo pipefail
           BASE="origin/${BASE_REF}"
+          # Directories considered critical: changes here that slip through a
+          # stale-base merge can break the site, the analysis pipeline, or CI.
+          CRITICAL_PATH_RE='^(src/data/|src/components/|scripts/cross-platform-verification/|scripts/spss/|scripts/minitab/|\.github/workflows/)'
           MERGE_BASE=$(git merge-base "$BASE" HEAD)
           PR_HEAD=$(git rev-parse HEAD)
           BASE_HEAD=$(git rev-parse "$BASE")
@@ -64,6 +71,11 @@ jobs:
 
           if [ "$MERGE_BASE" = "$BASE_HEAD" ]; then
             echo "Branch is up to date with $BASE_REF -- no phantom revert possible."
+            {
+              echo "## ✅ Phantom Revert Guard: clean"
+              echo ""
+              echo "Branch is up to date with \`$BASE_REF\` — no phantom-revert risk."
+            } >> "$GITHUB_STEP_SUMMARY"
             echo "violations=" >> "$GITHUB_OUTPUT"
             exit 0
           fi
@@ -78,10 +90,14 @@ jobs:
           # Files modified by this PR.
           PR_TOUCHED=$(git diff --name-only "$MERGE_BASE" HEAD | sort -u)
 
+          # Helper: strip blank lines so wc -l and comm see 0 lines for an empty
+          # variable rather than 1 (echo/printf always emit a trailing newline).
+          strip_blank() { printf '%s\n' "$1" | sed '/^$/d'; }
+
           # Use printf+sed to strip blank lines before counting so wc -l
           # reports 0 for an empty variable instead of 1 (echo adds a newline).
-          BASE_COUNT=$(printf '%s\n' "$BASE_TOUCHED" | sed '/^$/d' | wc -l)
-          PR_COUNT=$(printf '%s\n' "$PR_TOUCHED" | sed '/^$/d' | wc -l)
+          BASE_COUNT=$(strip_blank "$BASE_TOUCHED" | wc -l)
+          PR_COUNT=$(strip_blank "$PR_TOUCHED" | wc -l)
           echo "base touched: $BASE_COUNT files"
           echo "PR touched:   $PR_COUNT files"
 
@@ -90,7 +106,11 @@ jobs:
           # *not* change -- which is correct. But if the PR is then squash- or
           # rebase-merged through a stale base, the PR's old version of those
           # files (from MERGE_BASE) wins, silently reverting base's updates.
-          PHANTOM_CANDIDATES=$(comm -23 <(echo "$BASE_TOUCHED") <(echo "$PR_TOUCHED") || true)
+          # Use strip_blank() to avoid the stray blank line that `printf` adds for
+          # empty variables, which would make comm see a phantom empty entry.
+          PHANTOM_CANDIDATES=$(comm -23 \
+            <(strip_blank "$BASE_TOUCHED") \
+            <(strip_blank "$PR_TOUCHED") || true)
 
           if [ -z "$PHANTOM_CANDIDATES" ]; then
             if [ "$BEHIND_COUNT" -gt 5 ]; then
@@ -107,6 +127,12 @@ jobs:
               exit 1
             fi
             echo "No phantom-revert candidates."
+            {
+              echo "## ✅ Phantom Revert Guard: clean"
+              echo ""
+              echo "Branch is $BEHIND_COUNT commits behind \`$BASE_REF\`."
+              echo "No phantom-revert candidates detected in critical paths."
+            } >> "$GITHUB_STEP_SUMMARY"
             echo "violations=" >> "$GITHUB_OUTPUT"
             exit 0
           fi
@@ -133,10 +159,12 @@ jobs:
           } > /tmp/report.md
           cat /tmp/report.md >> "$GITHUB_STEP_SUMMARY"
 
-          # Hard-fail only if the branch is meaningfully behind. A handful of
-          # untouched files between branch and base is normal; dozens of files
-          # in critical paths are the symptom we caught on 2026-05-03.
-          CRITICAL_PATHS_HIT=$(echo "$PHANTOM_CANDIDATES" | grep -E '^(src/data/|src/components/|scripts/cross-platform-verification/|scripts/spss/|scripts/minitab/|\.github/workflows/)' || true)
+          # Hard-fail if any phantom candidate falls in a critical path, or if the
+          # branch is more than 5 commits behind -- either condition alone is
+          # enough to flag the PR as risky. A handful of untouched files in
+          # non-critical paths with a fresh branch is acceptable; critical-path
+          # hits or a very stale branch are the symptoms we caught on 2026-05-03.
+          CRITICAL_PATHS_HIT=$(echo "$PHANTOM_CANDIDATES" | grep -E "$CRITICAL_PATH_RE" || true)
 
           if [ -n "$CRITICAL_PATHS_HIT" ] || [ "$BEHIND_COUNT" -gt 5 ]; then
             echo ""
--- a/.github/workflows/phantom-revert-guard.yml
+++ b/.github/workflows/phantom-revert-guard.yml
@@ -97,10 +97,12 @@ jobs:
         run: |
           # Validate BASE_REF matches a safe ref pattern before passing to git.
           # Prevents shell injection from crafted workflow_dispatch inputs.
-          # Dots are allowed for version-style branches (e.g. fix/v1.2) but
-          # ".." sequences are explicitly blocked to prevent ref-range traversal.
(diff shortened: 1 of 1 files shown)


clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: This repo’s workflows consistently use `actions/checkout@v6` (e.g., `.github/workflows/ci.yml`) and set `FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true` at the workflow level. Consider updating this workflow to match those conventions to keep Actions runtime behavior consistent.
2. `.github/workflows/phantom-revert-guard.yml`: `pull-requests: write` permission appears unused in this workflow (no PR comments/labels/edits). Consider dropping it (least-privilege) to reduce token scope.

3. `.github/workflows/phantom-revert-guard.yml`: The workflow description says it hard-fails when the branch is >5 commits behind, but the script exits early when `PHANTOM_CANDIDATES` is empty, so `BEHIND_COUNT` is never computed/checked in that case. If the intent is to enforce the behind threshold unconditionally, move the behind-count check before the `PHANTOM_CANDIDATES` empty early-exit (and consider still writing a summary).
4. `.github/workflows/phantom-revert-guard.yml`: The critical-path regex uses `.github/workflows/` with an unescaped leading dot, so `grep -E` will treat `.` as “any character” and may match unintended paths. Escape it (e.g., `^\.(github/workflows/)` or `^\.github/workflows/`) to ensure only the `.github/workflows/` directory is treated as critical.

5. `.github/workflows/phantom-revert-guard.yml`: Typo in comment: “dozens or files” → “dozens of files”.

6. `.github/workflows/phantom-revert-guard.yml`: `wc -l` on an empty `BASE_TOUCHED`/`PR_TOUCHED` string will report `1` (because `echo` emits a trailing newline), so these file counts can be off by one. Consider using a newline-safe count (e.g., `printf '%s\n' ... | sed '/^$/d' | wc -l`) or storing the file lists in arrays.


clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: `comm` is fed via `echo "$BASE_TOUCHED"` / `echo "$PR_TOUCHED"`. When either variable is empty, `echo` still emits a blank line, which can make `PHANTOM_CANDIDATES` contain a stray newline and bypass the `-z` check (false positive phantom candidates). Use `printf` + blank-line stripping (like the wc -l logic) or avoid variables and run `comm` directly on the `git diff | sort -u` process substitutions.

2. `.github/workflows/phantom-revert-guard.yml`: The PR description says this guard “always writes a categorized report to the workflow step summary”, but on the up-to-date early exit path nothing is appended to `$GITHUB_STEP_SUMMARY`. Consider writing a short “no issues found” summary here (and in the other success early-exit) so the workflow output is consistent.
3. `.github/workflows/phantom-revert-guard.yml`: This workflow only runs on `pull_request` events tied to the PR head SHA. If `main` advances while a PR is open (no new commits pushed), this job will not rerun automatically, so a branch can become stale without the guard re-evaluating. If the intent is to catch staleness as it happens, consider adding a mechanism to re-run on base-branch updates (e.g., a `push` on the base branch that dispatches checks for open PRs, or a scheduled run).
4. `.github/workflows/phantom-revert-guard.yml`: The header comment claims “A regular CI run is happy because it builds B in isolation”, but this repo’s CI workflow checks out the PR merge commit by default (actions/checkout without a head SHA ref), so it’s not strictly building the branch in isolation. Tweaking this explanation would help avoid confusion about what additional risk this guard is covering.
5. `.github/workflows/phantom-revert-guard.yml`: The comment says “Hard-fail only if the branch is meaningfully behind”, but the condition also hard-fails on *any* `CRITICAL_PATHS_HIT` regardless of `BEHIND_COUNT`. Either adjust the comment to match the behavior, or change the condition if the behind-threshold is intended to gate failures.

6. `.github/workflows/phantom-revert-guard.yml`: On the “no phantom-revert candidates” success path, nothing is written to `$GITHUB_STEP_SUMMARY` even though the PR description states the workflow always writes a categorized report. Consider appending a brief “no candidates / behind count” summary here as well so every run leaves an audit trail in the summary UI.

clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: The comment says critical paths cover changes that could break “the analysis pipeline”, but `CRITICAL_PATH_RE` does not include `scripts/analysis/` (which is where the analysis pipeline lives, e.g. validate-analysis.yml targets `scripts/analysis/**`). As-is, phantom-revert candidates under `scripts/analysis/` won’t trigger the critical-path failure condition. Consider adding `scripts/analysis/` to the regex (and PR description’s list), or adjust the comment to match the intended scope.


clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: `actions/checkout` is pinned to `ref: github.event.pull_request.head.sha` but does not set `repository:` to the PR’s head repo. For PRs opened from forks, that SHA may not exist in the base repo’s `origin`, so checkout/fetch can fail and the guard won’t run. Either restrict this workflow to same-repo PRs (like `copilot-review-cycle.yml` does) or configure checkout to pull from `github.event.pull_request.head.repo.full_name` and add a separate remote/fetch for the base repo.
2. `.github/workflows/phantom-revert-guard.yml`: The failure condition is `CRITICAL_PATHS_HIT` OR `BEHIND_COUNT > 5`, but the error message always claims both (“behind … and has untouched updates in critical paths”). This will be misleading when only one condition is true. Consider emitting separate, accurate messages (or composing the message based on which condition(s) triggered).


clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: `workflow_dispatch` runs this job, but the steps rely on `github.event.pull_request.*` (checkout `ref`, fetch base ref, and `BASE_REF` env). For manual dispatch events these fields are undefined, so the workflow will fail (e.g., checking out an empty ref / fetching an empty base branch). Either remove `workflow_dispatch`, or add `workflow_dispatch` inputs (e.g., base ref + PR head SHA/branch) and branch the logic so dispatch uses `${{ github.sha }}` and a defined base ref (often `${{ github.event.repository.default_branch }}`) instead of `github.event.pull_request.*`.

clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: `steps.refs.outputs.base_ref` can come from `workflow_dispatch` inputs and is interpolated unquoted into `git fetch origin ...`. A crafted value containing shell metacharacters (e.g., `main; ...`) would be executed by bash. Quote the ref and use `--` (and ideally validate it matches an expected ref/sha pattern) before passing it to git.

clarkemoyer:
**Copilot Review Cycle (Round 7/7):** Max rounds reached. 3 comment(s) still need human attention.

clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: For `workflow_dispatch`, `base_ref` can reasonably be provided as a fully-qualified ref like `refs/heads/main`. That value passes the current validation but later gets used as `origin/${BASE_REF}`, which would become `origin/refs/heads/main` and fail `git rev-parse`/`git merge-base`. Consider normalizing inputs (e.g., strip a leading `refs/heads/`) or tightening validation to accept only branch names so manual runs don’t break unexpectedly.
2. `.github/workflows/phantom-revert-guard.yml`: `git fetch origin -- "$BASE_REF"` doesn’t guarantee that a remote-tracking ref like `origin/$BASE_REF` exists in all cases (especially from detached-SHA checkouts). Since the next step relies on `BASE="origin/${BASE_REF}"`, fetch with an explicit refspec mapping (e.g., `$BASE_REF:refs/remotes/origin/$BASE_REF`) to ensure the reference is created/updated deterministically (this pattern is already used in other workflows).

3. `.github/workflows/phantom-revert-guard.yml`: The PR description’s “critical path” list doesn’t mention `scripts/analysis/`, but the workflow treats it as critical via `CRITICAL_PATH_RE`. Either update the PR description to include it or remove it from the regex so the documented behavior matches the implemented failure conditions.

clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: The base_ref input validation still allows values starting with '-' (e.g. '-n'), which can be interpreted by `git fetch` as an option rather than a refspec. Tighten validation (e.g., require first char alnum, or use `git check-ref-format --branch`) before calling `git fetch` so workflow_dispatch inputs can’t trigger unintended git behavior.

2. `.github/workflows/phantom-revert-guard.yml`: `comm` errors are currently being suppressed via `|| true` when computing PHANTOM_CANDIDATES. If the inputs ever become unsorted (or another real error occurs), the command will fail and this will silently treat it as “no candidates,” potentially missing phantom reverts. Prefer removing `|| true` (or explicitly handling only the specific non-error status you expect) so failures are surfaced.


clarkemoyer:

1. `.github/workflows/phantom-revert-guard.yml`: The BASE_REF validation regex is stricter than Git’s actual branch-name rules (e.g., underscores are valid in refnames but rejected here). This can cause the workflow to fail for legitimate base branches or workflow_dispatch inputs. Consider validating with `git check-ref-format --branch "$BASE_REF"` (and/or applying the validation only for workflow_dispatch inputs) so all valid refs pass while still preventing option-style/injection cases.

2. `.github/workflows/phantom-revert-guard.yml`: This comment is misleading: the workflow exits early in the up-to-date branch case before computing BEHIND_COUNT, so it’s not actually computed “before any early-exit”. Suggest rewording to reflect that BEHIND_COUNT is computed early in the stale-branch path (or move the computation above the up-to-date early return).


clarkemoyer:

|---|---|---|---|---|
| 1 | 6 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 01:13 |
| 2 | 6 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 01:29 |
| 3 | 1 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 01:44 |
| 4 | 2 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 01:55 |
| 5 | 1 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 02:07 |
| 6 | 1 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 02:21 |
| 7 | 3 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 02:34 |
| 8 | 3 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 04:10 |
| 9 | 2 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 04:23 |
| 10 | 2 | phantom-revert-guard.yml | 🔧 Fixed | 2026-05-04 04:36 |
| 11 | 0 | — | ✅ Clean | 2026-05-04 04:51 |

### Round 1 Comments (6)

- `.github/workflows/phantom-revert-guard.yml` — This repo’s workflows consistently use `actions/checkout@v6` (e.g., `.github/workflows/ci.yml`) and set `FORCE_JAVASCRIP
- `.github/workflows/phantom-revert-guard.yml` — `pull-requests: write` permission appears unused in this workflow (no PR comments/labels/edits). Consider dropping it (l
- `.github/workflows/phantom-revert-guard.yml` — The workflow description says it hard-fails when the branch is >5 commits behind, but the script exits early when `PHANT
- `.github/workflows/phantom-revert-guard.yml` — The critical-path regex uses `.github/workflows/` with an unescaped leading dot, so `grep -E` will treat `.` as “any cha
- `.github/workflows/phantom-revert-guard.yml` — Typo in comment: “dozens or files” → “dozens of files”. 
- `.github/workflows/phantom-revert-guard.yml` — `wc -l` on an empty `BASE_TOUCHED`/`PR_TOUCHED` string will report `1` (because `echo` emits a trailing newline), so the

### Round 2 Comments (6)

- `.github/workflows/phantom-revert-guard.yml` — `comm` is fed via `echo "$BASE_TOUCHED"` / `echo "$PR_TOUCHED"`. When either variable is empty, `echo` still emits a bla
- `.github/workflows/phantom-revert-guard.yml` — The PR description says this guard “always writes a categorized report to the workflow step summary”, but on the up-to-d
- `.github/workflows/phantom-revert-guard.yml` — This workflow only runs on `pull_request` events tied to the PR head SHA. If `main` advances while a PR is open (no new 
- `.github/workflows/phantom-revert-guard.yml` — The header comment claims “A regular CI run is happy because it builds B in isolation”, but this repo’s CI workflow chec
- `.github/workflows/phantom-revert-guard.yml` — The comment says “Hard-fail only if the branch is meaningfully behind”, but the condition also hard-fails on *any* `CRIT
- `.github/workflows/phantom-revert-guard.yml` — On the “no phantom-revert candidates” success path, nothing is written to `$GITHUB_STEP_SUMMARY` even though the PR desc

### Round 3 Comments (1)

- `.github/workflows/phantom-revert-guard.yml` — The comment says critical paths cover changes that could break “the analysis pipeline”, but `CRITICAL_PATH_RE` does not 

### Round 4 Comments (2)

- `.github/workflows/phantom-revert-guard.yml` — `actions/checkout` is pinned to `ref: github.event.pull_request.head.sha` but does not set `repository:` to the PR’s hea
- `.github/workflows/phantom-revert-guard.yml` — The failure condition is `CRITICAL_PATHS_HIT` OR `BEHIND_COUNT > 5`, but the error message always claims both (“behind …

### Round 5 Comments (1)

- `.github/workflows/phantom-revert-guard.yml` — `workflow_dispatch` runs this job, but the steps rely on `github.event.pull_request.*` (checkout `ref`, fetch base ref, 

### Round 6 Comments (1)

- `.github/workflows/phantom-revert-guard.yml` — `steps.refs.outputs.base_ref` can come from `workflow_dispatch` inputs and is interpolated unquoted into `git fetch orig

### Round 7 Comments (3)

- `.github/workflows/phantom-revert-guard.yml` — Job-level `if:` references `github.event.pull_request.*` without guarding on `event_name == 'pull_request'`. For `workfl
- `.github/workflows/phantom-revert-guard.yml` — `git fetch origin -- "$BASE_REF"` may not update/create the expected `refs/remotes/origin/$BASE_REF` (especially when `a
- `.github/workflows/phantom-revert-guard.yml` — The PR description’s “critical path” list does not include `scripts/analysis/`, but `CRITICAL_PATH_RE` treats it as crit

### Round 8 Comments (3)

- `.github/workflows/phantom-revert-guard.yml` — For `workflow_dispatch`, `base_ref` can reasonably be provided as a fully-qualified ref like `refs/heads/main`. That val
- `.github/workflows/phantom-revert-guard.yml` — `git fetch origin -- "$BASE_REF"` doesn’t guarantee that a remote-tracking ref like `origin/$BASE_REF` exists in all cas
- `.github/workflows/phantom-revert-guard.yml` — The PR description’s “critical path” list doesn’t mention `scripts/analysis/`, but the workflow treats it as critical vi

### Round 9 Comments (2)

- `.github/workflows/phantom-revert-guard.yml` — The base_ref input validation still allows values starting with '-' (e.g. '-n'), which can be interpreted by `git fetch`
- `.github/workflows/phantom-revert-guard.yml` — `comm` errors are currently being suppressed via `|| true` when computing PHANTOM_CANDIDATES. If the inputs ever become 

### Round 10 Comments (2)

- `.github/workflows/phantom-revert-guard.yml` — The BASE_REF validation regex is stricter than Git’s actual branch-name rules (e.g., underscores are valid in refnames b
- `.github/workflows/phantom-revert-guard.yml` — This comment is misleading: the workflow exits early in the up-to-date branch case before computing BEHIND_COUNT, so it’

### Result

✅ All automated review comments addressed in 4 round(s). Ready for final human review.


**Copilot Review Cycle:** ✅ Passed after 4 round(s) - review clean, all CI checks green. Ready to merge.

PR #1909, opened 2026-05-06
fix(prolific): switch message export to per-user fetch (no 30-day limit)

clarkemoyer:
## Summary

The first run of [prolific-message-export.yml](.github/workflows/prolific-message-export.yml) on main failed with HTTP 400:

> This API can only retrieve the last 30 days of messages.

The `/messages/?created_after=` endpoint has a 30-day cap that I didn't see documented when I built the workflow.

## Fix

Switch to the per-user endpoint (`/messages/?user_id=`), which has no time limit:

1. Enumerate every participant in the study's submission roster (`prolific_list_studies` + `prolific_submissions`, both already in `tabs_api.py`).
2. Fetch per-participant via `prolific_user_messages`.
3. Deduplicate by message id.
4. Apply the `SINCE` env-var filter post-fetch so the operator can still scope a run without re-pulling.

Trades 1 paginated request for ~600 individual requests (one per unique participant). This is the only path to the full history.

## Test plan

- [x] Script parses.
- [ ] CI passes.
- [ ] After merge, `gh workflow run prolific-message-export.yml` produces all 5 artifacts.

### DIFF
--- a/scripts/analysis/extract_prolific_messages.py
+++ b/scripts/analysis/extract_prolific_messages.py
@@ -72,7 +72,11 @@ def _check_outside_repo(env_var: str, path: Path) -> None:
 
 # Re-use the project's Prolific helpers.
 sys.path.insert(0, str(REPO_ROOT / "scripts" / "analysis"))
-from tabs_api import prolific_recent_messages  # noqa: E402
+from tabs_api import (  # noqa: E402
+    prolific_list_studies,
+    prolific_submissions,
+    prolific_user_messages,
+)
 
 # ---------------------------------------------------------------------------
 # PII risk heuristics (matches q74_full_analysis.py)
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `scripts/analysis/extract_prolific_messages.py`: When STUDY_ID is provided, the per-participant fetch currently pulls *all* messages for each user and never filters them back down to the requested study. This diverges from the previous behavior (study-filtered fetch) and can include unrelated study conversations in the artifacts. After collecting/deduping messages, filter by (m.get('data') or {}).get('study_id') == study_id when study_id is set.
2. `scripts/analysis/extract_prolific_messages.py`: The error log prints the full participant/user id. These IDs are sensitive and will end up in GitHub Actions logs; please redact (e.g., prefix/suffix only) or avoid logging the raw id altogether.
3. `scripts/analysis/extract_prolific_messages.py`: Exceptions from prolific_user_messages() are swallowed and the script continues, which can silently produce an incomplete export while still exiting successfully. Track failures and fail the run (non-zero exit) or implement retry/backoff so operators can trust the artifact completeness.
4. `scripts/analysis/extract_prolific_messages.py`: This change introduces new control flow (study enumeration, per-participant fetch, dedupe, since post-filter) but this script has no direct test coverage. Consider adding pytest coverage that mocks tabs_api helpers to validate: (1) study_id filtering, (2) dedupe by message id across participants, and (3) failure handling behavior.

PR #1910, opened 2026-05-06
fix(prolific): handle datetime_created field + write raw before since-filter

clarkemoyer:
## Summary

Run 25455092532 succeeded but produced an empty artifact. Logs:

> After study_id filter: 1032 messages (dropped 241 from other studies)
> After since-filter (2025-01-01T00:00:00Z): 0 messages (dropped 1032)

## Root cause

The `/messages/?user_id=` endpoint returns timestamps in **`datetime_created`**, not `sent_at` / `created_at` like `/messages/?created_after=` does. My since-filter only checked the latter two, treated every message as untimed (`""` < `since`), and dropped them all.

## Fix

1. Module-level `_msg_ts()` helper checks all three fields. Used everywhere we read a timestamp (filter, sort, transcripts, CSV) so the format inconsistency only lives in one place.
2. Untimed messages now pass through the filter (over-include in the local PII review file rather than silently drop).
3. Raw dump is written **before** the since-filter so the operator can always inspect the unfiltered API response.
4. `::warning::` line if the since-filter drops everything — points the operator at the raw dump.

## Test plan

- [x] All 10 existing tests still pass.
- [ ] CI passes.
- [ ] After merge, re-dispatch produces non-empty artifact (~1032 messages expected).

### DIFF
--- a/scripts/analysis/extract_prolific_messages.py
+++ b/scripts/analysis/extract_prolific_messages.py
@@ -339,18 +339,20 @@ def main() -> int:
     # always inspect the unfiltered API response if a downstream filter
     # over-prunes (it has previously - the per-user endpoint returns
     # timestamps under different field names than /messages/?created_after).
+    pre_since_count = len(messages)
     with raw_path.open("w", encoding="utf-8") as f:
         json.dump(messages, f, indent=2, default=str)
-    print(f"Raw dump (pre-since): {raw_path}")
+    print(f"Raw dump (pre-since): {raw_path}  [{pre_since_count} messages]")
 
     # Optional post-fetch filter: keep only messages on or after `since`.
     # `_msg_ts` is module-level so the sort/transcript/csv code below
     # uses the same logic. If a message has no timestamp at all we KEEP
     # it (better to over-include in the local PII review file than to
-    # silently drop everything).
+    # silently drop everything). The walrus operator ensures _msg_ts() is
+    # called only once per message.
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `scripts/analysis/extract_prolific_messages.py`: The since-filter calls `_msg_ts(m)` twice per message (`not _msg_ts(m) or _msg_ts(m) >= since`), which repeats lookups and makes the condition harder to read. Compute the timestamp once per message (e.g., in a small helper/local variable or via an assignment expression) and then apply the filter.

2. `scripts/analysis/extract_prolific_messages.py`: `print(f"API returned {len(messages)} messages total.")` is now misleading because `messages` may have been reduced by the since-filter, while the raw dump written just above is pre-since. Consider logging both counts (pre-since and post-since) or updating the wording to reflect that this is the post-filter count.
3. `scripts/analysis/extract_prolific_messages.py`: The new timestamp normalization in `_msg_ts()` (including `datetime_created`) and the updated since-filter behavior (keeping untimed messages) aren’t covered by existing tests. Add unit tests that verify: (1) messages with only `datetime_created` are kept/dropped correctly by `SINCE`, and (2) messages with no timestamp fields pass through the since-filter as intended.

clarkemoyer:
**Copilot Review Cycle (Round 2/7):** Timed out waiting for Copilot review. You can re-trigger manually.

Issue #1900, opened 2026-05-06
Platform-by-the-Numbers: canonical, auto-updating LOC + commits + workflows + integrations metric


clarkemoyer:
## Problem

The defense slides, CRP body, and website all carry slightly different (and sometimes wrong) figures for "how much code is in this project." Specifically, defense Slide 5 advertises:

> 4,680+ Commits | 574K+ Lines of Code | 39 GitHub Actions | 8+ Integrations | 5-Phase Development

Three of these five numbers are stale or wrong as of today's main:

| Metric | Slide 5 claim | Actual today (cc12a560) | Method |
|---|---|---|---|
| Commits | 4,680+ | **5,087** | `git rev-list --count HEAD` |
| Lines of Code | **574K+** | **149,687** code-only / **237K total** raw | pygount (cloc-equivalent) |
| GitHub Actions | 39 | **47** | `ls .github/workflows/*.yml \| wc -l` |
| Integrations | 8+ | **8** (cloudflare, github, google-analytics, google-jules, microsoft-clarity, prolific, qualtrics, zotero) | from `src/app/making-of-tabs/integrations/` |
| Development Phases | 5-Phase | 5 (narrative; CRP §3.6 Table 8) | descriptive constant |

The CRP body has its own numbers ("190,000+ lines", "120,000+ core application", "217,141 total lines", "8,921 lines pipeline") that are similarly drifting.

The 574K LOC figure on the slide is the worst offender — it is ~3.8× the canonical cloc-counted number and likely double-counted node_modules or generated files. This is a defense-day risk.

## Why a manually-maintained number won't work

Every metric here changes weekly. A static "574K+" on a slide goes stale immediately and disagrees with the CRP and the live repo. The fix is to compute these in CI and surface them through a JSON the website + future presentation materials can fetch.

## Industry standard for "lines of code"

The de-facto LOC counters and what they report:

- **cloc** (Perl, ~16 years old, citation standard) - reports `blank | comment | code` separately, by language, with auto-skip for binaries. Used by Linux Foundation, OpenSSF, academic papers.
- **scc** - Go, faster, drop-in compatible with cloc, also reports cyclomatic complexity & estimated COCOMO cost.
- **tokei** - Rust, fastest, similar output.
- **GitHub Linguist** - what powers GitHub's language bar; reports BYTES not lines (this is why GitHub doesn't show LOC natively).

The four numbers any serious project reports (not one):
[...]

PR #2142, opened 2026-05-09

clarkemoyer:

### DIFF
--- a/public/data/platform-stats.json
+++ b/public/data/platform-stats.json
@@ -1,7 +1,7 @@
 {
-  "as_of": "2026-05-09T12:54:18Z",
-  "commit_sha": "4add785cc72a88d5504f2766fbfa1665bcbac444",
-  "commit_count": 5755,
+  "as_of": "2026-05-09T12:58:29Z",
+  "commit_sha": "b4a14a46fc8eca9341e7363684a8ba484cffda9f",
+  "commit_count": 5757,
   "lines_of_code": 227595,
   "comments": 12848,
   "blanks": 26129,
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a full lowercase Git SHA, which contains 24+ consecutive `[0-9a-f]` characters and will be flagged by `scripts/check-data-pii.py` (the PII guard runs on every PR that touches `src/data/**`). Convert the stored SHA to a format that cannot match `[0-9a-f]{24}` (e.g., uppercase the hex, shorten it to <24 chars, or split with a non-hex separator), and update the generator workflow accordingly so the next refresh doesn’t reintroduce the violation.

2. `public/data/platform-stats.json`: This public mirror should use the same `commit_sha` representation as `src/data/platform-stats.json`. If you change the SHA formatting there to avoid the PII guard’s `[0-9a-f]{24}` match (e.g., uppercase/shorten/split), apply the identical change here so `/data/platform-stats.json` stays consistent with the source-of-truth stats used at build time.


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a full lowercase Git SHA, which contains 24+ consecutive `[0-9a-f]` characters and will be flagged by `scripts/check-data-pii.py` (the PII guard runs on every PR that touches `src/data/**`). Convert the stored SHA to a format that cannot match `[0-9a-f]{24}` (e.g., uppercase the hex, shorten it to <24 chars, or split with a non-hex separator), and update the generator workflow accordingly so the next refresh doesn’t reintroduce the violation.

2. `public/data/platform-stats.json`: This public mirror should use the same `commit_sha` representation as `src/data/platform-stats.json`. If you change the SHA formatting there to avoid the PII guard’s `[0-9a-f]{24}` match (e.g., uppercase/shorten/split), apply the identical change here so `/data/platform-stats.json` stays consistent with the source-of-truth stats used at build time.


clarkemoyer:

PR #2233, opened 2026-05-11

clarkemoyer:

### DIFF
--- a/public/data/platform-stats.json
+++ b/public/data/platform-stats.json
@@ -1,7 +1,7 @@
 {
-  "as_of": "2026-05-11T02:45:08Z",
-  "commit_sha": "126fa2cda2fef4381ecb2bb76b10a490c850f1e4",
-  "commit_count": 5935,
+  "as_of": "2026-05-11T02:49:10Z",
+  "commit_sha": "5e08bf59752f2dcb33bdacb59c5bfc6116233ab1",
+  "commit_count": 5937,
   "lines_of_code": 227599,
   "comments": 12848,
   "blanks": 26129,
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-char lowercase hex string, which will be matched by the PII guard regex `[0-9a-f]{24}` (it finds 24-char substrings inside longer strings). This will cause `scripts/check-data-pii.py` / `Data Quality Check` to fail on any PR/push touching `src/data/**`. Consider storing an uppercase SHA, a shortened SHA, or inserting separators, and update the generator workflow accordingly so future refreshes don’t reintroduce the violation.

2. `public/data/platform-stats.json`: Mirror file also contains a 40-char lowercase `commit_sha` that matches the `[0-9a-f]{24}` PID pattern used for data PII scanning. Even if the scan only targets `src/data/**`, keeping this value uppercase/shortened (to match the sanitized source file) avoids propagating PID-like substrings into publicly served assets.


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-char lowercase hex string, which will be matched by the PII guard regex `[0-9a-f]{24}` (it finds 24-char substrings inside longer strings). This will cause `scripts/check-data-pii.py` / `Data Quality Check` to fail on any PR/push touching `src/data/**`. Consider storing an uppercase SHA, a shortened SHA, or inserting separators, and update the generator workflow accordingly so future refreshes don’t reintroduce the violation.

2. `public/data/platform-stats.json`: Mirror file also contains a 40-char lowercase `commit_sha` that matches the `[0-9a-f]{24}` PID pattern used for data PII scanning. Even if the scan only targets `src/data/**`, keeping this value uppercase/shortened (to match the sanitized source file) avoids propagating PID-like substrings into publicly served assets.


clarkemoyer:

PR #2333, opened 2026-05-11

clarkemoyer:

### DIFF
--- a/public/data/platform-stats.json
+++ b/public/data/platform-stats.json
@@ -1,7 +1,7 @@
 {
-  "as_of": "2026-05-11T09:33:39Z",
-  "commit_sha": "30133c853d5af0c0a62f5b443da46eb7949ea13e",
-  "commit_count": 6133,
+  "as_of": "2026-05-11T09:37:41Z",
+  "commit_sha": "6b94743792d4f867fd2957a1c1ee05db11b3631f",
+  "commit_count": 6135,
   "lines_of_code": 227599,
   "comments": 12848,
   "blanks": 26129,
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-char lowercase hex string, which contains a `[0-9a-f]{24}` substring and will be flagged by `scripts/check-data-pii.py` (the PII scan runs on all `src/data/**/*.json`). This will cause CI to fail even though the value is a git SHA. Consider storing an uppercase SHA (or otherwise breaking the 24-lowercase-hex run, e.g. short SHA) and updating the generator workflow to emit the same format so it doesn’t regress on the next refresh.

2. `public/data/platform-stats.json`: This mirror file includes the same lowercase 40-char `commit_sha`, which matches the project’s PID regex (`[0-9a-f]{24}`) used by the `src/data` PII guard. Once `src/data/platform-stats.json` switches to an uppercase/short/broken-up SHA format, this public mirror should be kept consistent so consumers don’t copy/paste a value that would trip the guard.


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-char lowercase hex string, which contains a `[0-9a-f]{24}` substring and will be flagged by `scripts/check-data-pii.py` (the PII scan runs on all `src/data/**/*.json`). This will cause CI to fail even though the value is a git SHA. Consider storing an uppercase SHA (or otherwise breaking the 24-lowercase-hex run, e.g. short SHA) and updating the generator workflow to emit the same format so it doesn’t regress on the next refresh.

2. `public/data/platform-stats.json`: This mirror file includes the same lowercase 40-char `commit_sha`, which matches the project’s PID regex (`[0-9a-f]{24}`) used by the `src/data` PII guard. Once `src/data/platform-stats.json` switches to an uppercase/short/broken-up SHA format, this public mirror should be kept consistent so consumers don’t copy/paste a value that would trip the guard.


clarkemoyer:

PR #2480, opened 2026-05-13

clarkemoyer:

### DIFF
--- a/public/data/platform-stats.json
+++ b/public/data/platform-stats.json
@@ -1,7 +1,7 @@
 {
-  "as_of": "2026-05-13T02:17:10Z",
-  "commit_sha": "29bf68b84a12cbf9adf8a876bf6f7da12e6814c0",
-  "commit_count": 6417,
+  "as_of": "2026-05-13T02:21:19Z",
+  "commit_sha": "bc8bb5a4411d4487ee2dde344dd048d2a9ac8858",
+  "commit_count": 6419,
   "lines_of_code": 227599,
   "comments": 12848,
   "blanks": 26129,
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-character lowercase hex string, which contains a 24-character lowercase hex substring and will be flagged by `scripts/check-data-pii.py` (regex `[0-9a-f]{24}`) when CI scans `src/data/**/*.json`. This will cause the Data Quality Check workflow to fail for this PR. Consider emitting an uppercase SHA (or a short SHA < 24 chars, or splitting the SHA with a delimiter) in the generator so it can’t match the PID pattern.

2. `public/data/platform-stats.json`: This public mirror also includes the full lowercase `commit_sha`. Even if the PII scan currently targets `src/data/**`, keeping this value uppercase/short (in sync with `src/data/platform-stats.json`) avoids future false positives if the scan scope expands and keeps the two copies consistent.


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a 40-character lowercase hex string, which contains a 24-character lowercase hex substring and will be flagged by `scripts/check-data-pii.py` (regex `[0-9a-f]{24}`) when CI scans `src/data/**/*.json`. This will cause the Data Quality Check workflow to fail for this PR. Consider emitting an uppercase SHA (or a short SHA < 24 chars, or splitting the SHA with a delimiter) in the generator so it can’t match the PID pattern.

2. `public/data/platform-stats.json`: This public mirror also includes the full lowercase `commit_sha`. Even if the PII scan currently targets `src/data/**`, keeping this value uppercase/short (in sync with `src/data/platform-stats.json`) avoids future false positives if the scan scope expands and keeps the two copies consistent.


clarkemoyer:

PR #2501, opened 2026-05-13

clarkemoyer:

### DIFF
--- a/public/data/platform-stats.json
+++ b/public/data/platform-stats.json
@@ -1,7 +1,7 @@
 {
-  "as_of": "2026-05-13T12:12:45Z",
-  "commit_sha": "1b792c88546dbdf0efa28d81ccf011d19bca9bbb",
-  "commit_count": 6453,
+  "as_of": "2026-05-13T12:17:22Z",
+  "commit_sha": "b21614fedfcaac256d336d21beecce59a6277965",
+  "commit_count": 6455,
   "lines_of_code": 227603,
   "comments": 12848,
   "blanks": 26129,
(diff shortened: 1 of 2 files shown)


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a full lowercase git SHA, which will match the PII guard regex (`[0-9a-f]{24}`) used by `scripts/check-data-pii.py` when scanning `src/data/**/*.json`. This will cause the Data Quality Check workflow to report a violation and block the PR. Consider uppercasing the SHA (or otherwise breaking up the lowercase-hex run) in this JSON, and update the generating workflow so future refreshes stay compliant.

2. `public/data/platform-stats.json`: This public mirror keeps `commit_sha` as a full lowercase git SHA. Even if the PII scan only targets `src/data/**`, it’s best to keep the public copy consistent with whatever transformation is applied there (e.g., uppercase SHA) so consumers don’t see two different identifiers and future tooling doesn’t inadvertently scan/flag this copy.


clarkemoyer:

1. `src/data/platform-stats.json`: `commit_sha` is a full lowercase git SHA, which will match the PII guard regex (`[0-9a-f]{24}`) used by `scripts/check-data-pii.py` when scanning `src/data/**/*.json`. This will cause the Data Quality Check workflow to report a violation and block the PR. Consider uppercasing the SHA (or otherwise breaking up the lowercase-hex run) in this JSON, and update the generating workflow so future refreshes stay compliant.

2. `public/data/platform-stats.json`: This public mirror keeps `commit_sha` as a full lowercase git SHA. Even if the PII scan only targets `src/data/**`, it’s best to keep the public copy consistent with whatever transformation is applied there (e.g., uppercase SHA) so consumers don’t see two different identifiers and future tooling doesn’t inadvertently scan/flag this copy.


clarkemoyer:

PR #2582, opened 2026-05-15
feat(ops): auto-dispatch stale request-returns and auto-reject stale RR window

clarkemoyer:
## Summary

Automates the daily manual stale-bucket workflow we've been running by hand. Two new pipeline phases slot in after Phase 3c (which already produces `stale-triage.json`), and Phase 5 reports what each one did.

- **Phase 3d (Auto-Request-Return)** — reads `stale_no_reply_to_message` bucket (TABS messaged >48h ago, no reply, no API request-return yet) and POSTs `/submissions/{id}/request-return/` per PID with disposition-aware reason text.
- **Phase 3e (Auto-Reject-Stale-RR)** — reads `stale_no_reply_to_rr` bucket (formal API request-return sent >48h ago, no reply) and rejects via existing `reject_by_pid.py` before Prolific's reserve timeout auto-approves them.

## Safety controls

- **Ceiling**: 30 PIDs per run by default (`vars.AUTO_STALE_MAX_PER_RUN`). Above the ceiling, the phase emits exit code 2 with a MANUAL INTERVENTION NEEDED banner in the daily report.
- **Kill switches**: `vars.AUTO_REQUEST_RETURN_ENABLED` and `vars.AUTO_REJECT_STALE_ENABLED` (both default `'true'`; set to `'false'` to pause that phase).
- **Privacy**: PIDs redacted to last-4 in the daily issue body per the 2026-04-23 post-mortem; full lists live in 7-day retention artifacts.
- **No surprise scope creep**: scripts refuse unmodelled dispositions and unknown PIDs; auto-reject runs with `STRICT_AWAITING=false` so a single out-of-band status change does not fail the whole batch.

## Files changed

| File | Change |
|---|---|
| `scripts/analysis/tabs_api.py` | new `prolific_request_return()` |
| `scripts/analysis/request_return_by_pid.py` | NEW disposition-aware script |
| `scripts/analysis/reject_by_pid.py` | adds `MAX_PER_RUN`, `STRICT_AWAITING`, `OUTPUT_JSON_PATH` |
| `scripts/analysis/tests/test_auto_stale_actions.py` | NEW (14 cases) |
| `.github/workflows/daily-pipeline.yml` | Phase 3d + 3e + Phase 5 deps |
| `scripts/build-daily-report.sh` | two new sections + critical-findings hook |

## Test plan

- [x] 14 new pytest cases pass locally (`pytest scripts/analysis/tests/test_auto_stale_actions.py`)
- [x] No new failures in the rest of the Python suite (10 pre-existing Windows-only `charmap` failures confirmed present on main)
- [x] Smoke-tested daily-report Python rendering against synthetic ceiling-exceeded + 3-success + 1-failure JSONs
- [x] `bash -n scripts/build-daily-report.sh` syntax check
- [x] YAML re-formatted with prettier
- [ ] CI green
- [ ] Workflow-dispatch a manual daily-pipeline run on this branch to verify Phase 3d/3e wire-up against real artifacts before merge

### DIFF
--- a/.github/workflows/daily-pipeline.yml
+++ b/.github/workflows/daily-pipeline.yml
@@ -619,6 +619,149 @@ jobs:
           path: ${{ runner.temp }}/stale-triage.json
           retention-days: 1
 
+  # ── Phase 3d: Auto-dispatch API request-return for stale-after-message PIDs ─
+  # Reads stale-triage.json's `stale_no_reply_to_message` bucket (TABS messaged
+  # > 48h ago, no participant reply, no API-level return request yet) and runs
+  # request_return_by_pid.py to dispatch /submissions/{id}/request-return/ per
+  # PID. Disposition-aware reason text per PID. Ceiling-guarded.
+  #
+  # Kill switch: vars.AUTO_REQUEST_RETURN_ENABLED (default 'true').
+  # Ceiling:     vars.AUTO_STALE_MAX_PER_RUN     (default '30').
+  auto-request-return:
+    name: 'Phase 3d: Auto-Request-Return'
+    needs: [analyze, generate-dashboard]
+    if: ${{ needs.generate-dashboard.result == 'success' && vars.AUTO_REQUEST_RETURN_ENABLED != 'false' }}
+    runs-on: ubuntu-latest
+    environment: prolific-prod
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+      - uses: actions/download-artifact@v8
+        with:
+          name: disposition-csv
+          path: ${{ runner.temp }}
+      - uses: actions/download-artifact@v8
+        with:
+          name: stale-triage
+          path: ${{ runner.temp }}
+
+      - name: Extract stale_no_reply_to_message PIDs
+        id: pids
+        run: |
+          set -euo pipefail
+          PID_LIST=$(python3 -c "
+          import json
+          with open('${{ runner.temp }}/stale-triage.json', encoding='utf-8') as f:
+              d = json.load(f)
+          pids = [r['pid'] for r in d.get('buckets', {}).get('stale_no_reply_to_message', []) if r.get('pid')]
+          print(','.join(pids))
+          ")
+          COUNT=$(python3 -c "print(len('$PID_LIST'.split(',')) if '$PID_LIST' else 0)")
+          echo "pid_list=$PID_LIST" >> "$GITHUB_OUTPUT"
+          echo "count=$COUNT" >> "$GITHUB_OUTPUT"
+          echo "Found $COUNT PIDs in stale_no_reply_to_message bucket"
+
+      - name: Auto request-return
+        if: ${{ steps.pids.outputs.count != '0' }}
+        run: python3 scripts/analysis/request_return_by_pid.py
+        env:
+          PROLIFIC_API_TOKEN: ${{ secrets.TABS_PROLIFIC_TOKEN }}
+          STUDY_ID: ${{ vars.PROLIFIC_STUDY_ID }}
+          CSV_FILE_PATH: ${{ runner.temp }}/disposition.csv
+          PID_LIST: ${{ steps.pids.outputs.pid_list }}
+          MAX_PER_RUN: ${{ vars.AUTO_STALE_MAX_PER_RUN || '30' }}
+          DRY_RUN: 'false'
+          CONFIRM_REQUEST_RETURN: 'REQUEST_RETURN'
+          OUTPUT_JSON_PATH: ${{ runner.temp }}/auto-request-return-results.json
+
+      - name: Write empty results JSON when no PIDs
+        if: ${{ steps.pids.outputs.count == '0' }}
+        run: |
+          cat > "${{ runner.temp }}/auto-request-return-results.json" <<EOF
+          {"mode": "no_action", "input_count": 0, "successes": [], "failures": [], "skipped_non_awaiting": [], "ceiling_exceeded": false}
+          EOF
+
+      - uses: actions/upload-artifact@v7
+        if: always()
+        with:
+          name: auto-request-return-results
+          path: ${{ runner.temp }}/auto-request-return-results.json
+          retention-days: 7
+
+  # ── Phase 3e: Auto-reject stale-no-reply-to-return-request ───────────────
+  # Reads stale-triage.json's `stale_no_reply_to_rr` bucket (formal API
+  # request-return sent > 48h ago, no participant reply since). Prolific
+  # auto-APPROVES these after the reserve timeout, so they MUST be rejected
+  # before that window closes. Ceiling-guarded.
+  #
+  # Kill switch: vars.AUTO_REJECT_STALE_ENABLED (default 'true').
+  # Ceiling:     vars.AUTO_STALE_MAX_PER_RUN   (default '30').
+  auto-reject-stale:
+    name: 'Phase 3e: Auto-Reject-Stale-RR'
+    needs: [analyze, generate-dashboard, auto-request-return]
+    if: ${{ always() && needs.generate-dashboard.result == 'success' && vars.AUTO_REJECT_STALE_ENABLED != 'false' }}
+    runs-on: ubuntu-latest
+    environment: prolific-prod
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+      - uses: actions/download-artifact@v8
+        with:
+          name: disposition-csv
+          path: ${{ runner.temp }}
+      - uses: actions/download-artifact@v8
+        with:
+          name: stale-triage
+          path: ${{ runner.temp }}
+
+      - name: Extract stale_no_reply_to_rr PIDs
+        id: pids
+        run: |
+          set -euo pipefail
+          PID_LIST=$(python3 -c "
+          import json
+          with open('${{ runner.temp }}/stale-triage.json', encoding='utf-8') as f:
+              d = json.load(f)
+          pids = [r['pid'] for r in d.get('buckets', {}).get('stale_no_reply_to_rr', []) if r.get('pid')]
+          print(','.join(pids))
+          ")
+          COUNT=$(python3 -c "print(len('$PID_LIST'.split(',')) if '$PID_LIST' else 0)")
+          echo "pid_list=$PID_LIST" >> "$GITHUB_OUTPUT"
+          echo "count=$COUNT" >> "$GITHUB_OUTPUT"
+          echo "Found $COUNT PIDs in stale_no_reply_to_rr bucket"
+
+      - name: Auto reject
+        if: ${{ steps.pids.outputs.count != '0' }}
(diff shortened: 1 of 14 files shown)


clarkemoyer:

1. `scripts/analysis/reject_by_pid.py`: The live auto-reject path only revalidates that a PID is still AWAITING REVIEW. A participant reply after `stale-triage.json` was generated does not change that status, so the unattended job can still reject someone who responded during the race window; re-run the stale/no-reply check (or verify message history after the triage timestamp) immediately before rejecting.
2. `scripts/analysis/request_return_by_pid.py`: The live request-return path only revalidates submission status. Participant replies and existing `return_requested` flags both leave the submission in AWAITING REVIEW, so an out-of-band reply or manual request-return after `stale-triage.json` is produced can still trigger another request-return notification.

3. `scripts/analysis/request_return_by_pid.py`: Invalid `MAX_PER_RUN` values are silently treated as no ceiling because non-digit input sets `max_per_run` to `None`. A typo in the repository variable would disable the safety cap for this live participant operation instead of failing closed or using the default.
4. `scripts/analysis/reject_by_pid.py`: Invalid `MAX_PER_RUN` values are silently treated as no ceiling because non-digit input sets `max_per_run` to `None`. A malformed repository variable would disable the safety cap for live rejections instead of failing closed or using the default.

5. `scripts/analysis/request_return_by_pid.py`: A target PID missing from the current Prolific submissions is treated as `skipped_non_awaiting` and the script can still exit successfully. That contradicts the safety property of refusing unknown PIDs and can hide a wrong-study/stale-artifact problem while leaving the participant unactioned; `NOT FOUND` should be a failure, not a benign skip.

6. `scripts/analysis/reject_by_pid.py`: With `STRICT_AWAITING=false`, a PID missing from the current Prolific submissions is recorded as a skipped non-awaiting status and does not fail the run. A `NOT FOUND` PID indicates the artifact or study mapping is wrong and should not be treated like an out-of-band status change, especially because the stale reject candidate will otherwise be silently left to auto-approve.
7. `scripts/build-daily-report.sh`: This message assumes any skipped auto phase was disabled by a repository variable, but these jobs also skip when their upstream dashboard dependency does not succeed. In those cases the report would tell operators the kill switch was used even though the phase never had valid inputs; distinguish disabled from upstream-skipped states before rendering this text.

8. `scripts/analysis/tabs_api.py`: There are existing unit tests for the other Prolific write helpers that assert the exact URL and JSON payload, but this new write helper is not covered. Please add a mocked `_http`/`_json_request` test for `prolific_request_return()` so endpoint drift or an incorrect `request_return_reasons` payload is caught before it affects real participants.
9. `.github/workflows/daily-pipeline.yml`: Adding an automatic rejection phase leaves the workflow's safety documentation stale: the header still says the daily pipeline “Does NOT reject anyone - all rejections require manual trigger.” Update that operator-facing documentation in the same change so reviewers do not rely on the old manual-only safety model.
10. `scripts/analysis/tests/test_auto_stale_actions.py`: This import is unused in the new test module. Removing it keeps the test file tidy and avoids future confusion about whether subprocess calls were intended to be mocked.

11. `scripts/analysis/request_return_by_pid.py`: On ceiling-exceeded runs this JSON omits the target PID list, while the report tells operators to recover it from the upstream `stale-triage` artifact (retained only 1 day). If nobody intervenes immediately, the 7-day auto-request-return artifact will not contain the full list needed for manual follow-up.
12. `scripts/analysis/reject_by_pid.py`: On ceiling-exceeded runs this JSON omits the target PID list, while the report points operators back to the upstream `stale-triage` artifact (retained only 1 day). After that expires, the 7-day auto-reject artifact will not contain the full list needed to prevent stale request-return candidates from auto-approving.

clarkemoyer:

1. `scripts/build-daily-report.sh`: The report still renders the pre-action stale triage section immediately before the new auto-action results, and that section tells operators to “Reject these” / “Dispatch request-return for these” for the same buckets the new phases may have already processed. This can make the daily issue instruct a duplicate manual action unless the stale section is rewritten or suppressed/annotated based on the Phase 3d/3e results.

2. `scripts/analysis/reject_by_pid.py`: This adds a tight per-PID call to `prolific_user_messages` without the delay/retry behavior that `find_stale_return_requested.py` uses for this same endpoint. A 30-PID batch can burst message-history requests and fail revalidation under Prolific rate limiting; reuse the existing backoff helper or add equivalent pacing before running this unattended.

This issue also appears on line 500 of the same file.
3. `scripts/analysis/request_return_by_pid.py`: For `FLAG-SMEAL`, the flagged duration is between `smealBenchmarkMin` and `smealBenchmarkMax` (5–9 minutes), and the existing participant message cites the 9-minute Smeal benchmark. This reason uses `SMEAL_BENCHMARK_MIN` (5 minutes), so a 6–8 minute completion is described as below a 5-minute benchmark, which is inaccurate in the user-facing Prolific return-request reason.

This issue also appears on line 383 of the same file.

clarkemoyer:
**Copilot Review Cycle (Round 2/7):** Fixes were pushed but CI is failing. Pausing review cycle until CI is fixed.

clarkemoyer:
## Real-data validation on dispatched run [25921525576](https://github.com/clarkemoyer/technologyadoptionbarriers.org/actions/runs/25921525576)

**Phase 3d (Auto-Request-Return): 7/7 success live** — today's daily increment had 7 fresh stale-no-reply-to-message PIDs that crossed 48h since this morning. The phase dispatched all 7 via the Prolific API. Confirmed in the artifact:
\`\`\`json
{"mode": "live", "input_count": 7, "successes": 7, "failures": 0, "ceiling": 30}
\`\`\`

**Phase 3e (Auto-Reject-Stale-RR): no_action** — zero PIDs in the \`stale_no_reply_to_rr\` bucket today, as expected (the 22 request-returns from this morning are still within their 48h windows).

**Report-rendering quirk:** Phase 5 of \`daily-pipeline.yml\` uses \`ref: main\` for the report-builder checkout (a deliberate stability pin — if someone dispatches with a buggy report builder, the issue still renders cleanly). So the test-run issue [#2584](https://github.com/clarkemoyer/technologyadoptionbarriers.org/issues/2584) is missing the new sections because it ran main's old \`build-daily-report.sh\`. **Sections will appear post-merge** when tomorrow's scheduled pipeline picks them up.

To prove the rendering works, I replayed \`build-daily-report.sh\` locally against run-25921525576's real artifacts — both sections render correctly with last-4 PID redaction, ceiling display, and proper success/failure tables.

Ready to merge once Copilot review clears.

clarkemoyer:

1. `scripts/analysis/request_return_by_pid.py`: This live revalidation loop calls the Prolific messages API once per PID without the backoff helper or the inter-call delay used by the other stale-action paths. At the default ceiling of 30 PIDs, transient errors or rate limiting can turn an otherwise safe unattended run into a phase failure; use the existing `_fetch_messages_with_backoff`/`_API_CALL_DELAY` pattern before each per-PID message fetch.
2. `scripts/analysis/request_return_by_pid.py`: This emits each full Prolific PID into the workflow log during the automated daily run. The PR's privacy control keeps full PID lists in the short-retention artifact and redacts the issue body, but workflow logs are a separate, longer-lived surface; redact these log lines to last-4 and keep the full values only in the JSON artifact.

This issue also appears on line 419 of the same file.
3. `.github/workflows/daily-pipeline.yml`: This new scheduled auto-reject phase invokes `reject_by_pid.py`, which prints full PIDs in its plan/live log output (for example `scripts/analysis/reject_by_pid.py:371` and `:536`). That creates a longer-lived workflow-log copy even though the privacy control says full PID lists should live in the short-retention artifact; add a redacted/nonverbose mode for the automated path or redact the script output before logging.

4. `scripts/build-daily-report.sh`: This label is inaccurate for several entries in the same result schema: both stale-action scripts put `NO_LONGER_ELIGIBLE:<bucket>` revalidation skips into `skipped_non_awaiting` even when the submission is still AWAITING REVIEW. Reporting all of them as “not AWAITING REVIEW” can mislead operators; use a neutral label such as “Skipped / no longer eligible” or split status skips from revalidation skips.


clarkemoyer:

1. `scripts/analysis/request_return_by_pid.py`: This live revalidation loop calls `prolific_user_messages` directly for each PID without the backoff/delay used elsewhere for stale-action message fetches. `find_stale_return_requested._fetch_messages_with_backoff` documents that callers must skip classification on fetch failure and applies retries plus `_API_CALL_DELAY`; without that here, transient Prolific rate limits/API glitches can incorrectly fail PIDs or the whole unattended phase.
2. `scripts/analysis/reject_by_pid.py`: This new revalidation is unconditional, so `reject_by_pid.py` no longer rejects a confirmed PID unless it currently classifies as `stale_no_reply_to_rr`. The existing manual `prolific-reject-by-pid.yml` workflow exposes this script as a general “Reject submissions by PID” tool and does not pass any override, so a live operator-initiated rejection for any other AWAITING REVIEW case will be skipped and can exit successfully with zero rejections. Consider gating this stale-bucket check behind a pipeline-specific option or updating the workflow/API contract explicitly.
3. `scripts/analysis/request_return_by_pid.py`: This logs full Prolific participant IDs during the automated daily request-return phase. The PR’s privacy model keeps full PID lists in short-retention artifacts and redacts the issue body; writing the same identifiers to workflow logs can expose them outside that artifact boundary and for a different retention/access period. Redact here as well and keep full IDs only in the JSON artifact.
4. `scripts/analysis/request_return_by_pid.py`: This live-action log line also prints the full PID. Since these automated workflow logs are not the short-retention results artifact described in the PR privacy controls, redact the PID in stdout/stderr and keep the full identifier only in the machine-readable artifact.
5. `scripts/build-daily-report.sh`: This label is inaccurate for several entries written to `skipped_non_awaiting`: both scripts also use that array for stale revalidation skips such as `NO_LONGER_ELIGIBLE:in_window_rr` or participant replies while the submission may still be AWAITING REVIEW. The report should use a broader label (for example, “Skipped / no longer eligible”) so operators do not misread these as status-only skips.

6. `scripts/build-daily-report.sh`: Failure reasons are inserted into a markdown table after escaping pipes, but newline/control characters from API error bodies are not normalized. A multiline Prolific error will break the table (and can hide following rows outside the details block); strip `\r`/`\n` before truncating, like the stale-triage renderer does for reasons.

7. `scripts/analysis/request_return_by_pid.py`: The live success path for the new request-return script is not covered by the added tests: current live tests only cover missing submissions and revalidation skips. Add a test where `classify_submission` returns `stale_no_reply_to_message` and assert `prolific_request_return` is called with the resolved submission ID and disposition-specific reason, because this is the core production side effect of Phase 3d.
8. `scripts/analysis/reject_by_pid.py`: The new live revalidation branch is not covered for the successful path: added tests cover ceiling/dry-run, missing PID, and no-longer-eligible skips, but not the case where `classify_submission` returns `stale_no_reply_to_rr` and the script actually calls `prolific_reject_submission`. Add that test so the core Phase 3e production side effect is protected.
9. `scripts/build-daily-report.sh`: The new auto-action jobs are only treated as critical when their result is exactly `failure`. If one is `cancelled` (for example due to timeout or a cancelled environment approval), the report can still be labeled auto-close-eligible even though the automated stale action did not complete. Treat non-success/non-skipped results for these phases as critical, while still allowing intentional kill-switch skips.

PR #2594, opened 2026-05-15
Update daily pipeline data

clarkemoyer:
Automated update from the unified daily pipeline.

**Data flow**: Qualtrics export → Prolific enrichment →
disposition waterfall → analysis suite → CRP dataset (N=200) →
approve CLEAN → dashboard summary.

All statistics computed across 5 sample definitions.
CRP dataset: N=200 tiered selection + NIST de-identification.

PR #1925, opened 2026-05-07
fix(ci): keep platform-stats in scan trigger and allow commit_sha keys

clarkemoyer:
## Why

`Data Quality Check` was failing on `src/data/platform-stats.json` because `scripts/check-data-pii.py` matches any 24-char lowercase hex substring, and `commit_sha` (40-char Git SHA) contains such substrings.

## Fix

1. Keep PR/push protection intact by scanning all `src/data/**` (including `platform-stats.json`).
2. Handle the false-positive category in the scanner by allowlisting commit SHA leaf keys:
   - `commit_sha`
   - `commitSha`

This preserves immediate PII enforcement on PRs/pushes while eliminating commit-hash false positives.

## Validation

- `python3 scripts/check-data-pii.py src/data/platform-stats.json` → no violations
- Synthetic test with `user_pid` still fails as expected
- `npm run lint`
- `npm test`
- `npm run build`

PR #2618, opened 2026-05-16
fix(test): scope response-funnel value assertions to their metric cards

clarkemoyer:
## Summary

Fixes the failing Jest test `Response Funnel page (QualtricsSurveyStats) › renders Qualtrics question IDs and items presented as separate values` that has been blocking the daily-pipeline data PR (#2594) from merging.

## Root cause

The test uses `screen.getByText(...)` (single-match) with exact-match values. It still fails any day a pipeline-driven number coincidentally equals an instrument constant. Today: `TOTAL_ITEMS_PRESENTED = 57` and `dispositions["FLAG-PARTIAL-STRAIGHTLINING"] = 57` both render in `<div class="mt-2 text-3xl font-bold text-slate-900">57</div>`, so `getByText("57")` finds two matches and throws.

The author's earlier switch to exact-match avoided substring collisions (e.g. `21` matching inside `421`) but did not address full-value collisions.

## Fix

Scope each assertion to its specific `MetricCard` via the card's `<h4>` heading, then use `within(card).getByText(value)`. This makes the assertion immune to other cards' values regardless of pipeline drift.

No production code changed — test-only.

## Test plan

- [ ] Jest `response-funnel.test.tsx` passes locally and in CI
- [ ] Other 5 tests in the same file still pass (unchanged)
- [ ] Once green, PR #2594 (daily pipeline data) can re-run checks and merge

## Notes

- Discovered while triaging why `src/data/disposition-summary.json` on `main` was ~26 hours stale.
- Today's pipeline ran successfully; data is sitting on `chore/daily-pipeline-data` (PR #2594), blocked by this test plus the PII-scan false-positive that #1925 fixes.

### DIFF
--- a/__tests__/components/survey-stats/response-funnel.test.tsx
+++ b/__tests__/components/survey-stats/response-funnel.test.tsx
@@ -50,10 +50,10 @@ describe('Response Funnel page (QualtricsSurveyStats)', () => {
       .getByRole('heading', { name: /^Items presented$/i })
       .closest('.rounded-lg') as HTMLElement
     expect(
-      within(questionIdsCard).getByText(metricsData.questionCount.toLocaleString()),
+      within(questionIdsCard).getByText(metricsData.questionCount.toLocaleString())
     ).toBeInTheDocument()
     expect(
-      within(itemsPresentedCard).getByText(TOTAL_ITEMS_PRESENTED.toLocaleString()),
+      within(itemsPresentedCard).getByText(TOTAL_ITEMS_PRESENTED.toLocaleString())
     ).toBeInTheDocument()
   })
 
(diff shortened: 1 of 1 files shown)


clarkemoyer:

1. `__tests__/components/survey-stats/response-funnel.test.tsx`: Avoid coupling the test to Tailwind/styling classes and masking potential nulls. `heading.closest('.rounded-lg') as HTMLElement` will silently accept `null` at compile time and will also break if the card’s class list changes; consider using a semantic container (e.g., `closest('section')`/parent traversal from the heading) and/or asserting the container is non-null before calling `within(...)` for clearer failures.

PR #2938, opened 2026-06-18
feat(prolific): add stale_hours override to reject-by-pid workflow

clarkemoyer:
Adds optional stale_hours input to prolific-reject-by-pid.yml, passed to reject_by_pid.py as STALE_HOURS (default 48, behavior unchanged). Operator escape hatch for deliberate human-decision rejects of items parked in human_review_high_tier whose Prolific return_requested timestamp is recent. Dry-run + live verified on 2 AUTO-EXCLUDE 3/3-IRI submissions.

### DIFF
--- a/.github/workflows/prolific-reject-by-pid.yml
+++ b/.github/workflows/prolific-reject-by-pid.yml
@@ -47,8 +47,8 @@ on:
       stale_hours:
         description: 'Override the post-return-request staleness threshold in hours (default 48). Lower it for a deliberate human-decision reject of items already past the real return window but whose Prolific return_requested timestamp is recent.'
         required: false
-        type: string
-        default: '48'
+        type: number
+        default: 48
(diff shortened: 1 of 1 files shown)


clarkemoyer:

1. `.github/workflows/prolific-reject-by-pid.yml`: `stale_hours` is defined as a `string`, but this repo already uses `type: number` for numeric `workflow_dispatch` inputs (e.g. `.github/workflows/tabs-pr-triage.yml`). Using `number` here prevents non-numeric values from reaching the script and improves the Actions UI validation.
2. `.github/workflows/prolific-reject-by-pid.yml`: To match other workflows and avoid passing an empty string if the input is cleared in the UI, consider adding a fallback when setting `STALE_HOURS`. This keeps the default behavior stable even if the operator deletes the prefilled value.

